PR #117, opened 2026-05-23
fix: make trigger label a tag field with multi-label support (closes #114)

### LINKED ISSUE #114: Make Label as tag and support additional labels
We have issue_labeled -"autopilot ready".In addition to this, we can support ai_pilot_ready and ai_ready. We can make the text field  as a tag field. allow 3 ("autopilot ready") -- add support for ai_pilot_ready and ai_ready
Screenshot as below

Image

sseshachala:
## Summary

Resolves #114 — **Make Label as tag and support additional labels**

#### Frontend (`apps/web`)
- **`src/lib/config-schemas.ts`**:
  - Added `"tags"` to the `ConfigField.type` union
  - Added optional `suggestions?: string[]` field to `ConfigField` interface
  - Changed the trigger's `config.label` field (single `text`) → `config.labels` (`tags` type)
  - Pre-populated suggestions: `autopilot ready`, `ai_pilot_ready`, `ai_ready`

- **`src/components/canvas/BlockEditor.tsx`**:
  - Added `TagInput` component — chip-based multi-tag input with keyboard support (Enter/comma to add, Backspace to remove)
  - Shows quick-add suggestion chips for the three well-known autopilot labels
  - Wired `tags` type into `FieldInput` renderer
  - Updated trigger field filter key from `config.label` → `config.labels`

#### Backend (`apps/api`)
- **`app/routers/webhooks.py`**:
  - Added `AUTOPILOT_LABELS` constant: `{"autopilot ready", "ai_pilot_ready", "ai_ready"}`
  - Updated `_labels_match()` to prioritise the new `labels` list field (comma-separated or YAML list) with `label_mode: one_of` (default) or `all_of`
  - Backward-compatible: old `label: "autopilot ready"` single-string config still works

- **`app/dsl/schema.py`**:
  - Added comprehensive docstring to `Trigger` model documenting both legacy single-label and new multi-label YAML syntax

#### Playbooks
- **`playbooks/autopilot.yaml`**: Updated `on:` trigger to use `labels: [autopilot ready, ai_pilot_ready, ai_ready]` with `label_mode: one_of`
- **`playbooks/ai_ready.yaml`**: Updated `on:` trigger to use `labels: [ai_ready, ai-ready]` with `label_mode: one_of`

### UI Behaviour
The trigger block editor now shows a tag-chip input instead of a plain text field. Users can:
- Type a label name and press **Enter** or **comma** to add it as a chip
- Click the **×** on any chip to remove it
- Click the dashed quick-add buttons below (`+ autopilot ready`, `+ ai_pilot_ready`, `+ ai_ready`) to add a well-known label in one click
- Backspace with empty input removes the last tag

### DIFF
--- a/apps/api/app/dsl/schema.py
+++ b/apps/api/app/dsl/schema.py
@@ -189,7 +189,34 @@ class Block(BaseModel):
 # DAG entry time.
 # ---------------------------------------------------------------------------
 class Trigger(BaseModel):
-    """One concrete trigger. Keyed in YAML by its event name (e.g. github.issue_labeled)."""
+    """
+    One concrete trigger. Keyed in YAML by its event name (e.g. github.issue_labeled).
+
+    Label matching for ``github.issue_labeled`` events
+    ---------------------------------------------------
+    Single label (legacy, still supported)::
+
+        on:
+          github.issue_labeled:
+            label: autopilot ready
+
+    Multiple labels — fires when **any** of the listed labels is added::
+
+        on:
+          github.issue_labeled:
+            labels:
+              - autopilot ready
+              - ai_pilot_ready
+              - ai_ready
+            label_mode: one_of   # default when `labels` list is present
+
+    All labels must be present on the issue::
+
+        on:
+          github.issue_labeled:
+            labels: [autopilot ready, ai_pilot_ready]
+            label_mode: all_of
+    """
 
     integration: str | None = None
     next: str | None = Field(
@@ -197,7 +224,7 @@ class Trigger(BaseModel):
         description="Entry block id. Defaults to the first key under blocks: if omitted.",
     )
 
-    model_config = ConfigDict(extra="allow")  # event-specific filter fields
+    model_config = ConfigDict(extra="allow")  # event-specific filter fields (label, labels, label_mode, etc.)
 
 
 # ---------------------------------------------------------------------------
--- a/apps/api/app/routers/webhooks.py
+++ b/apps/api/app/routers/webhooks.py
@@ -388,25 +388,52 @@ def _parse_string_list(raw: Any) -> list[str]:
     return []
 
 
-def _labels_match(config: dict[str, Any], incoming_label: str, issue_labels: list[str], strict: bool) -> bool:
-    mode = str(config.get("label_mode") or "exact").strip()
+# Well-known autopilot labels — used as the default tag set when no explicit
+# label configuration is present.
+AUTOPILOT_LABELS = {"autopilot ready", "ai_pilot_ready", "ai_ready"}
 
-    if mode == "exact":
+
+def _labels_match(config: dict[str, Any], incoming_label: str, issue_labels: list[str], strict: bool) -> bool:
+    """
+    Determine whether the incoming GitHub label (or the issue's full label set)
+    satisfies the workflow trigger's label configuration.
+
+    Trigger config shapes (in priority order):
+      1. ``label_mode: one_of``  + ``labels: [...]``  → any label in the list matches
+      2. ``label_mode: all_of``  + ``labels: [...]``  → every label must be present
+      3. ``label_mode: exact``   + ``label: "<str>"`` → exact single-label match (legacy)
+      4. ``labels: [...]`` (no label_mode)            → implicit one_of
+      5. ``label: "<str>"`` (no label_mode)           → legacy exact match
+
+    When no label configuration is present in strict mode the trigger does NOT fire.
+    In permissive mode an unconfigured trigger fires on any label.
+    """
+    mode = str(config.get("label_mode") or "").strip()
+
+    # ── new multi-label config (config.labels list / comma-separated string) ──
+    configured_labels = _parse_string_list(config.get("labels"))
+    if configured_labels:
+        effective_mode = mode if mode in ("one_of", "all_of") else "one_of"
+        if effective_mode == "one_of":
+            return incoming_label in configured_labels or any(lbl in configured_labels for lbl in issue_labels)
+        if effective_mode == "all_of":
+            issue_set = set(issue_labels)
+            return all(lbl in issue_set for lbl in configured_labels)
+
+    # ── legacy single-label config (config.label string) ──
+    if mode == "exact" or not mode:
         required = str(config.get("label") or "").strip()
         if not required:
             return not strict
         return incoming_label == required
 
-    configured_labels = _parse_string_list(config.get("labels"))
-    if not configured_labels:
+    if mode == "one_of":
+        # one_of with no labels list — treat as unconfigured
         return not strict
 
-    if mode == "one_of":
-        return incoming_label in configured_labels or any(lbl in configured_labels for lbl in issue_labels)
-
     if mode == "all_of":
-        issue_set = set(issue_labels)
-        return all(lbl in issue_set for lbl in configured_labels)
+        # all_of with no labels list — treat as unconfigured
+        return not strict
 
     # Unknown mode -> fail closed in strict mode
     return not strict
--- a/apps/api/playbooks/ai_ready.yaml
+++ b/apps/api/playbooks/ai_ready.yaml
@@ -40,7 +40,11 @@ params:
 on:
   github.issue_labeled:
     integration: github
-    label: ai-ready
+    # Fire on any of these labels — mirrors the autopilot.yaml pattern
+    labels:
+      - ai_ready
+      - ai-ready
+    label_mode: one_of
     next: fetch_issue
 
 blocks:
--- a/apps/api/playbooks/autopilot.yaml
+++ b/apps/api/playbooks/autopilot.yaml
@@ -17,7 +17,12 @@ description: >
 on:
   github.issue_labeled:
     integration: github
-    label: autopilot ready
+    # Fire on any of these labels — add or remove labels as needed
+    labels:
+      - autopilot ready
+      - ai_pilot_ready
+      - ai_ready
+    label_mode: one_of
     next: fetch_issue
 
 blocks:
--- a/apps/web/src/components/canvas/BlockEditor.tsx
+++ b/apps/web/src/components/canvas/BlockEditor.tsx
@@ -161,6 +161,97 @@ function setNestedValue(obj: Record<string, unknown>, path: string, value: unkno
   return result
 }
 
+
+// ── Tag input ─────────────────────────────────────────────────────────────────
+
+const KNOWN_LABEL_SUGGESTIONS = ["autopilot ready", "ai_pilot_ready", "ai_ready"]
+
+function TagInput({
+  value,
+  suggestions,
+  placeholder,
+  onChange,
+}: {
+  value: string[]
+  suggestions?: string[]
+  placeholder?: string
+  onChange: (tags: string[]) => void
+}) {
+  const [inputVal, setInputVal] = useState("")
+  const allSuggestions = suggestions ?? KNOWN_LABEL_SUGGESTIONS
+
+  function addTag(tag: string) {
+    const trimmed = tag.trim()
+    if (trimmed && !value.includes(trimmed)) {
+      onChange([...value, trimmed])
+    }
+    setInputVal("")
+  }
+
+  function removeTag(tag: string) {
+    onChange(value.filter(t => t !== tag))
+  }
+
+  function handleKeyDown(e: React.KeyboardEvent<HTMLInputElement>) {
+    if (e.key === "Enter" || e.key === ",") {
+      e.preventDefault()
+      addTag(inputVal)
+    } else if (e.key === "Backspace" && !inputVal && value.length > 0) {
+      removeTag(value[value.length - 1])
+    }
+  }
+
+  const unusedSuggestions = allSuggestions.filter(s => !value.includes(s))
+
+  return (
+    <div className="space-y-2">
+      {/* Tag chips display */}
+      <div className="flex flex-wrap gap-1.5 min-h-[2rem] w-full border border-stone-200 rounded-lg px-2.5 py-1.5 bg-white focus-within:ring-2 focus-within:ring-indigo-200">
+        {value.map(tag => (
+          <span
+            key={tag}
+            className="inline-flex items-center gap-1 rounded-md bg-indigo-50 border border-indigo-200 px-2 py-0.5 text-xs font-medium text-indigo-700"
+          >
+            {tag}
+            <button
+              type="button"
+              onClick={() => removeTag(tag)}
+              className="text-indigo-400 hover:text-indigo-700 leading-none"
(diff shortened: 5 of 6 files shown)


sseshachala (commented):
## PR Review: `fix: make trigger label a tag field with multi-label support` (#117)

Overall this is a well-structured, well-documented feature PR. The logic is backward-compatible and the frontend TagInput component is solid. A few issues to address before merging:

---

### 🔴 Critical

**1. `_labels_match` — unreachable code path when `label_mode` is set but `labels` is empty**

In the new code flow, if a user writes:
```yaml
label_mode: one_of
label: "autopilot ready"  # NOTE: singular `label`, not `labels`
```
The function hits:
```python
configured_labels = _parse_string_list(config.get("labels"))  # → []
if configured_labels:  # False — skips new block
    ...
if mode == "exact" or not mode:  # mode = "one_of" → False
    ...
if mode == "one_of":              # True → returns `not strict`
    return not strict
```
Result: **the trigger silently fails or always fires** (depending on `strict`) instead of falling through to the legacy single-label path. A user with `label_mode: one_of` + `label: "..."` (singular) gets completely broken behavior with no error. This is a regression vs. the old code where `one_of` without `labels` was also broken, but it's now harder to detect because the legacy `label` field is still advertised as supported.

**Fix:** Before the mode-only branches at lines 430–436, also attempt the legacy `config.label` lookup, or add a validation error when `label_mode` is set without a matching `labels` list.

---

### 🟡 Warning

**2. `AUTOPILOT_LABELS` constant is defined but never used**

```python
AUTOPILOT_LABELS = {"autopilot ready", "ai_pilot_ready", "ai_ready"}
```
This constant was added in `webhooks.py` (line 393) with a comment suggesting it should be a "default tag set when no explicit label configuration is present," but it is **never referenced** anywhere in the module. Either wire it into the permissive-mode fallback logic as intended, or remove it to avoid confusion.

**3. `all_of` semantics may fire unexpectedly**

In the `all_of` branch:
```python
if effective_mode == "all_of":
    issue_set = set(issue_labels)
    return all(lbl in issue_set for lbl in configured_labels)
```
This checks whether all configured labels are present on the issue, but it does **not** verify that `incoming_label` itself is in `configured_labels`. This means the trigger fires on **any** label addition event as long as the required set happens to already be on the issue — not just when the last required label is being added. This could cause duplicate workflow runs. Consider adding `incoming_label in configured_labels` as a guard.

**4. Frontend: `TagInput` suggestions are hardcoded at module level, not driven by schema**

`KNOWN_LABEL_SUGGESTIONS` in `BlockEditor.tsx` is a static constant (`["autopilot ready", "ai_pilot_ready", "ai_ready"]`) used as the default when no `suggestions` prop is passed. This duplicates the suggestions already defined in `config-schemas.ts` and creates a maintenance inconsistency — a future change to the schema suggestions won't automatically update the component fallback.

**Fix:** Remove `KNOWN_LABEL_SUGGESTIONS` from `BlockEditor.tsx` and always require the `suggestions` prop to be passed from the schema (or make the fallback an empty array).

---

### 🔵 Suggestion

**5. No input sanitization / length cap on tag values**

The `addTag` function only trims whitespace. A user could paste an arbitrarily long string or a string containing special YAML characters (`:`, `[`, `]`, `{`, `}`) as a label name. Since tags are serialized back into YAML playbooks, consider:
- Adding a max-length guard (e.g., 100 chars)
- Stripping or warning on characters that break YAML inline sequences

**6. `onBlur` fires `addTag` even when user presses Enter**

```tsx
onBlur={() => { if (inputVal.trim()) addTag(inputVal) }}
```
When the user presses **Enter**, `addTag` is called, `inputVal` is cleared to `""`, and then if the input loses focus it calls `addTag("")` — which is safely guarded by the `trimmed &&` check. This is fine in practice, but if `addTag` had side effects the double-call pattern could be a footgun. Low severity, but worth a note.

**7. Schema doc comment lists `label_mode: exact` as supported, but `exact` mode is now only reachable via the legacy path without `label_mode` being set**

The docstring in `schema.py` lists:
```
3. label_mode: exact + label: "<str>" → exact single-label match (legacy)
```
But in `_labels_match`, `mode == "exact"` is only reached when `configured_labels` is empty **and** `mode == "exact"`. The docstring is technically accurate but could mislead users into thinking `label_mode: exact` is a first-class supported value. Consider deprecating it explicitly or adding a validator warning.

---

### ✅ What looks good

- Backward-compatibility with `label: "..."` single-label syntax is preserved ✔
- `_parse_string_list` handles both YAML list and comma-separated string gracefully ✔
- TagInput keyboard UX (Enter/comma to add, Backspace to remove last) is clean ✔
- `aria-label` on remove buttons is good for accessibility ✔
- Playbook updates (`autopilot.yaml`, `ai_ready.yaml`) are consistent with the new schema ✔
- The `config-schemas.ts` `suggestions` field is properly typed and optional ✔

sseshachala (commented):
## 🔍 Automated Code Review — PR #117

**`fix: make trigger label a tag field with multi-label support (closes #114)`**

---

### 🔴 Critical

**1. `_labels_match` — logic bug: `all_of` no longer checks `incoming_label`**
`apps/api/app/routers/webhooks.py`

In the refactored code the `all_of` branch only checks whether every configured label appears in `issue_labels`:
```python
return all(lbl in issue_set for lbl in configured_labels)
```
The webhook payload delivers **only the single label that was just added** as `incoming_label`; the full `issue_labels` list may not yet reflect that label if it is fetched before GitHub propagates the event. The original code had the same gap, but the new `all_of` semantics make this more impactful — a user configuring `all_of` expects the trigger to fire only when every label is present, but that evaluation is entirely dependent on how `issue_labels` is populated upstream. There is no guard or comment explaining this assumption. At minimum, assert/log a warning when `issue_labels` is empty and `all_of` is configured, or document the required upstream contract.

---

**2. `_labels_match` — silent fallthrough: `mode == "exact"` is now dead code**
`apps/api/app/routers/webhooks.py`

Before the refactor, `mode` defaulted to `"exact"` when not set. After the refactor, `mode` defaults to `""`. The new guard `if mode == "exact" or not mode:` is reached only after the `if configured_labels:` block returns early. This means a trigger with `label_mode: exact` **and** a `labels` list will be evaluated as `one_of` — ignoring the explicit `exact` mode. The intended priority is stated in the docstring but is not enforced in code. A caller can be surprised by this silent override.

---

### ⚠️ Warning

**3. `TagInput` — comma in tag value is silently consumed**
`apps/web/src/components/canvas/BlockEditor.tsx`

The `handleKeyDown` handler intercepts the comma key (`","`) and calls `addTag`, which commits whatever is currently in `inputVal`. If a user tries to type a label that legitimately contains a comma (unusual but possible), the comma will never appear in the tag. More importantly, `onBlur` calls `addTag(inputVal)` **without** splitting on commas first; but `handleKeyDown` also never splits the full string if the user pastes text like `"label1,label2"`. Consider calling `inputVal.split(",")` in `addTag` or `onBlur` to handle paste scenarios consistently.

---

**4. Backward-compat regression: existing `label_mode: one_of` config with no `labels` list now returns `not strict`**
`apps/api/app/routers/webhooks.py`

Before this PR, a trigger configured as:
```yaml
label_mode: one_of
label: "autopilot ready"
```
…would fall into the `one_of` branch and match. After the refactor, `configured_labels` is empty (no `labels` key), so the code skips straight to the legacy block. There `mode == "one_of"` is truthy, so it returns `not strict` — effectively disabling the trigger in strict mode. This is a silent behavior change for any existing playbooks using `label_mode: one_of` with the old single `label:` field.

---

**5. `AUTOPILOT_LABELS` constant is defined but never used**
`apps/api/app/routers/webhooks.py`

```python
AUTOPILOT_LABELS = {"autopilot ready", "ai_pilot_ready", "ai_ready"}
```
This set is introduced but not referenced anywhere in the module. It will confuse maintainers into thinking there is a default-fallback behavior when there is none. Either wire it into `_labels_match` as an intended default, or remove it.

---

**6. Missing `required: true` enforcement for the new `tags` field type**
`apps/web/src/components/canvas/BlockEditor.tsx`

The `FieldInput` renderer for `type === "tags"` has no visual or functional indicator for `required` fields (the `field.required` prop is ignored). All other field types (text, select, toggle, etc.) presumably surface a required indicator or validation. Omitting it means a user can save a trigger block with an empty labels array, which will cause a runtime error when the webhook fires.

---

### 💡 Suggestion

**7. `TagInput` — duplicate-tag feedback is silent**
`apps/web/src/components/canvas/BlockEditor.tsx`

When a user types a label that already exists in `value`, `addTag` silently resets `inputVal` without any feedback. Consider showing a brief inline message ("already added") or briefly highlighting the existing chip.

---

**8. `_parse_string_list` — no test for the new call-site with `config.get("labels")`**
`apps/api/app/routers/webhooks.py`

The helper is now called in two places. The unit tests (if any) for `_labels_match` should be updated to cover: (a) `labels` as a YAML list, (b) `labels` as a comma-separated string, (c) mixed `label_mode` values, and (d) the `all_of` with empty `issue_labels` edge-case. No tests are included in this PR.

---

**9. Schema docstring — `label_mode: one_of` is listed as the *default*, but the code comment says "implicit one_of"**
`apps/api/app/dsl/schema.py`

Minor inconsistency: the docstring says `label_mode: one_of   # default when labels list is present` while the code comment says `# implicit one_of`. Align language across docs and code to avoid confusion.

---

**10. `config-schemas.ts` — `suggestions` field typed as `string[]` but not validated at runtime**
`apps/web/src/lib/config-schemas.ts`

`ConfigField.suggestions` is a TypeScript-only interface field. There is no runtime guard in `FieldInput` or `TagInput` ensuring `suggestions` is an array before passing it to `allSuggestions`. If a misconfigured schema passes `suggestions: null`, `filter()` on `null` will throw. Add a defensive fallback: `const allSuggestions = Array.isArray(suggestions) ? suggestions : KNOWN_LABEL_SUGGESTIONS`.

---

### Summary

| Severity | Count |
|---|---|
| 🔴 Critical | 2 |
| ⚠️ Warning | 4 |
| 💡 Suggestion | 4 |
| **Total** | **10** |

The core logic changes in `webhooks.py` carry the highest risk — specifically the silent `all_of`/`exact` mode interactions and the unused `AUTOPILOT_LABELS` constant that may signal incomplete implementation. The frontend changes are generally clean but need required-field validation for the new `tags` type.

sseshachala (commented):
## Code Review — PR #117: Multi-label trigger support

Overall this is a well-structured, backward-compatible feature. The logic refactor in `_labels_match()` is clean and the frontend `TagInput` component is solid. A few issues worth addressing before merge.

---

### 🔴 Critical

**1. `AUTOPILOT_LABELS` constant is defined but never used (`webhooks.py`)**
The set `AUTOPILOT_LABELS = {"autopilot ready", "ai_pilot_ready", "ai_ready"}` is declared (with a comment implying it's a "default tag set when no explicit label configuration is present") but is never referenced anywhere in the function logic. Either wire it into the fallback path of `_labels_match()` (e.g., use it when `not strict` and no config is present) or remove it to avoid dead code that misleads future readers.

```python
# Added but never used:
AUTOPILOT_LABELS = {"autopilot ready", "ai_pilot_ready", "ai_ready"}
```

**2. Silent `all_of` early-return with no `return` path (`webhooks.py`)**
In the new `_labels_match()`, after the `configured_labels` block:
```python
if effective_mode == "one_of":
    return ...
if effective_mode == "all_of":
    return ...
# ← falls through here if effective_mode is somehow neither (impossible today,
#   but fragile — the function will fall through to the legacy path)
```
Since `effective_mode` is always `"one_of"` or `"all_of"`, this is technically safe today, but adding an explicit `else: return not strict` after the block would make the control flow unambiguous and prevent regressions if `effective_mode` values expand.

---

### ⚠️ Warning

**3. `label_mode: exact` no longer has any effect when `labels` list is present (`webhooks.py`)**
If a user writes:
```yaml
labels: [autopilot ready]
label_mode: exact
```
`exact` is not in `("one_of", "all_of")` so it silently falls back to `one_of`. This is a silent behavior change that could surprise users who explicitly set `label_mode: exact` alongside a `labels` list. Either document this in the docstring or add `exact` as a valid `effective_mode` for the labels path.

**4. Data persistence mismatch — `config.labels` vs `config.label` (`config-schemas.ts` / `BlockEditor.tsx`)**
The schema key was changed from `config.label` → `config.labels`. Any persisted workflow YAML or saved canvas state that contains `config.label: "autopilot ready"` (single string) will silently stop matching in the UI editor after this change — the field will render as empty since the key no longer exists. The backend handles this gracefully via the legacy fallback, but the frontend does not re-map `config.label` → `config.labels` on load. Consider adding a migration/coercion in `FieldInput` or in the data loading path to read `config.label` as a single-element array when `config.labels` is absent.

**5. `onBlur` fires after `onKeyDown` — duplicate tag insertion possible (`BlockEditor.tsx`)**
```tsx
onBlur={() => { if (inputVal.trim()) addTag(inputVal) }}
```
When the user presses **Enter** or **comma**, `handleKeyDown` calls `addTag()` and clears `inputVal`. But if the Enter key dismisses focus (e.g., pressing Enter in some form contexts), the `onBlur` fires after `handleKeyDown` with the original value still in `inputVal` state before React flushes — this can cause duplicate tag insertion in certain browser environments. A safer pattern is to track a ref flag or check the event sequence. At minimum add a note in the code about this risk.

---

### 💡 Suggestion

**6. `TagInput` suggestions list is duplicated in two places**
`KNOWN_LABEL_SUGGESTIONS` in `BlockEditor.tsx` and `suggestions: ["autopilot ready", "ai_pilot_ready", "ai_ready"]` in `config-schemas.ts` are identical. Consider exporting a single source-of-truth constant (e.g. from `config-schemas.ts`) and importing it into `BlockEditor.tsx` to avoid drift.

**7. `label_mode` field is not exposed in the UI**
The backend supports `label_mode: one_of | all_of` but the frontend tag field has no corresponding control. Users relying on `all_of` semantics must edit YAML directly. Even a simple toggle/select for "Match any / Match all" would complete the feature for GUI-driven workflows.

**8. Missing test coverage for the refactored `_labels_match()` paths**
The PR adds non-trivial control-flow changes to `_labels_match()`. No test file changes are included. At minimum, unit tests for:
- `labels` list + `one_of` (match and no-match)
- `labels` list + `all_of` (partial and full match)
- legacy `label:` single string
- empty/unconfigured strict vs. permissive

would significantly reduce regression risk.

---

### ✅ What looks good
- Backward compatibility for `label: "..."` single-string triggers is preserved.
- The `TagInput` keyboard UX (Enter/comma to add, Backspace to remove) is well-implemented.
- `_parse_string_list()` reuse for both comma-separated strings and YAML lists is a clean approach.
- The `Trigger` docstring additions are thorough and accurate.
- Playbook YAML updates are consistent and well-commented.

PR #127, opened 2026-05-23
fix: multi-label trigger support — closes #121 and #114

### LINKED ISSUE #114: Make Label as tag and support additional labels

### LINKED ISSUE #121: fix: address PR #117 review findings — _labels_match regression and unused constant
## Context
PR #117 (multi-label support / TagInput) was reviewed by the AI PR Reviewer. These issues must be fixed before merging.

### 1. `_labels_match` silent failure when `label_mode` set but `label:` (singular) used instead of `labels:`
In `apps/api/app/routers/webhooks.py`, the function `_labels_match` has a broken code path:
- User writes `label_mode: one_of` + `label: "autopilot ready"` (singular)
- `configured_labels = _parse_string_list(config.get("labels"))` returns `[]`
- Falls through to `if mode == "one_of": return not strict` — trigger silently always fires or never fires

**Fix:** Before the mode-only branches, also attempt the legacy `config.label` lookup when `configured_labels` is empty. Or raise a validation error when `label_mode` is set without a matching `labels` list.

### 2. `AUTOPILOT_LABELS` constant defined but never used
In `apps/api/app/routers/webhooks.py` (around line 393):
```python
AUTOPILOT_LABELS = {"autopilot ready", "ai_pilot_ready", "ai_ready"}
```
This is never referenced. Either wire it into the permissive-mode fallback or remove it.

### 3. `all_of` mode fires on any label event if required labels already present
The `all_of` branch does not verify that `incoming_label` is in `configured_labels`. This can cause duplicate runs. Add `incoming_label in configured_labels` as a guard before the `all()` check.

### 4. `KNOWN_LABEL_SUGGESTIONS` hardcoded in `BlockEditor.tsx`
In `apps/web/src/components/canvas/BlockEditor.tsx`, `KNOWN_LABEL_SUGGESTIONS` duplicates suggestions already in `config-schemas.ts`. Remove it and fall back to an empty array instead.

## Files to change
- `apps/api/app/routers/webhooks.py` — fixes 1, 2, 3
- `apps/web/src/components/canvas/BlockEditor.tsx` — fix 4

sseshachala:
## Summary

- Fixes `_labels_match` regression (issue #121): checks `config.labels` list first, then falls back to legacy `config.label` — eliminates the silent failure when `label_mode: one_of` is set but only `label:` (singular) is present
- `all_of` mode: adds `incoming_label in configured_labels` guard to prevent duplicate runs when unrelated labels are added
- Removes unused `AUTOPILOT_LABELS` constant
- `config-schemas.ts`: adds `tags` field type with `suggestions` support; swaps `config.label` (text) → `config.labels` (tags)  
- `BlockEditor.tsx`: adds `TagInput` chip component with keyboard support (Enter/comma to add, Backspace to remove); quick-add suggestion chips from `field.suggestions`; no hardcoded constant
- Playbooks (`autopilot`, `autopilot-quick`, `autopilot-approved`, `ai_ready`): all updated to `labels:` list + `label_mode: one_of`

## Test plan

- [ ] Trigger block in canvas shows chip-based label input with autopilot ready / ai_pilot_ready / ai_ready suggestions
- [ ] Adding/removing chips works (keyboard + click)
- [ ] Existing workflow with `config.label: "autopilot ready"` still fires (legacy fallback)
- [ ] `label_mode: one_of` + `labels: [...]` fires on any matching label
- [ ] `label_mode: all_of` + `labels: [a, b]` only fires when incoming label is in the list AND all labels are present on the issue
- [ ] PR #117 (old branch deleted) superseded by this PR — close #117

Closes #121, closes #114

### DIFF
--- a/apps/api/app/routers/webhooks.py
+++ b/apps/api/app/routers/webhooks.py
@@ -389,27 +389,25 @@ def _parse_string_list(raw: Any) -> list[str]:
 
 
 def _labels_match(config: dict[str, Any], incoming_label: str, issue_labels: list[str], strict: bool) -> bool:
-    mode = str(config.get("label_mode") or "exact").strip()
-
-    if mode == "exact":
-        required = str(config.get("label") or "").strip()
-        if not required:
-            return not strict
-        return incoming_label == required
-
+    mode = str(config.get("label_mode") or "").strip()
     configured_labels = _parse_string_list(config.get("labels"))
-    if not configured_labels:
+
+    if configured_labels:
+        effective_mode = mode if mode in ("one_of", "all_of") else "one_of"
+        if effective_mode == "one_of":
+            return incoming_label in configured_labels or any(lbl in configured_labels for lbl in issue_labels)
+        # all_of: only fire when the triggering label is one we care about
+        if effective_mode == "all_of":
+            if incoming_label not in configured_labels:
+                return False
+            issue_set = set(issue_labels)
+            return all(lbl in issue_set for lbl in configured_labels)
+
+    # Legacy / fallback — handles label: singular and label_mode: one_of with no labels list
+    required = str(config.get("label") or "").strip()
+    if not required:
         return not strict
-
-    if mode == "one_of":
-        return incoming_label in configured_labels or any(lbl in configured_labels for lbl in issue_labels)
-
-    if mode == "all_of":
-        issue_set = set(issue_labels)
-        return all(lbl in issue_set for lbl in configured_labels)
-
-    # Unknown mode -> fail closed in strict mode
-    return not strict
+    return incoming_label == required
 
 
 def _repo_matches(config: dict[str, Any], incoming_repo: str, strict: bool) -> bool:
--- a/apps/api/playbooks/ai_ready.yaml
+++ b/apps/api/playbooks/ai_ready.yaml
@@ -40,7 +40,10 @@ params:
 on:
   github.issue_labeled:
     integration: github
-    label: ai-ready
+    labels:
+      - ai_ready
+      - ai-ready
+    label_mode: one_of
     next: fetch_issue
 
 blocks:
--- a/apps/api/playbooks/autopilot-approved.yaml
+++ b/apps/api/playbooks/autopilot-approved.yaml
@@ -7,7 +7,11 @@ description: >
 on:
   github.issue_labeled:
     integration: github
-    label: autopilot ready
+    labels:
+      - autopilot ready
+      - ai_pilot_ready
+      - ai_ready
+    label_mode: one_of
     next: fetch_issue
 
 blocks:
--- a/apps/api/playbooks/autopilot-quick.yaml
+++ b/apps/api/playbooks/autopilot-quick.yaml
@@ -7,7 +7,11 @@ description: >
 on:
   github.issue_labeled:
     integration: github
-    label: autopilot ready
+    labels:
+      - autopilot ready
+      - ai_pilot_ready
+      - ai_ready
+    label_mode: one_of
     next: fetch_issue
 
 blocks:
--- a/apps/api/playbooks/autopilot.yaml
+++ b/apps/api/playbooks/autopilot.yaml
@@ -17,7 +17,11 @@ description: >
 on:
   github.issue_labeled:
     integration: github
-    label: autopilot ready
+    labels:
+      - autopilot ready
+      - ai_pilot_ready
+      - ai_ready
+    label_mode: one_of
     next: fetch_issue
 
 blocks:
--- a/apps/web/src/components/canvas/BlockEditor.tsx
+++ b/apps/web/src/components/canvas/BlockEditor.tsx
@@ -161,6 +161,73 @@ function setNestedValue(obj: Record<string, unknown>, path: string, value: unkno
   return result
 }
 
+// ── Tag input ─────────────────────────────────────────────────────────────────
+
+function TagInput({
+  value,
+  suggestions,
+  placeholder,
+  onChange,
+}: {
+  value: string[]
+  suggestions?: string[]
+  placeholder?: string
+  onChange: (tags: string[]) => void
+}) {
+  const [inputVal, setInputVal] = useState("")
+  const unusedSuggestions = (suggestions ?? []).filter(s => !value.includes(s))
+
+  function addTag(tag: string) {
+    const trimmed = tag.trim()
+    if (trimmed && !value.includes(trimmed)) onChange([...value, trimmed])
+    setInputVal("")
+  }
+
+  function removeTag(tag: string) {
+    onChange(value.filter(t => t !== tag))
+  }
+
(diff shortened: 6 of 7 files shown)


sseshachala (commented):
## 🤖 Automated Code Review — PR #127

**`fix: multi-label trigger support — closes #121 and #114`**

Overall the PR is well-scoped and the logic is sound. A few issues are worth addressing before merge.

---

### 🔴 Critical

**1. Silent regression: `label_mode: exact` is no longer supported (`webhooks.py`)**
The original code defaulted `mode` to `"exact"` and had a dedicated `exact` branch. The new code defaults `mode` to `""` (empty string) and silently falls through to `"one_of"` for any unknown mode value — including `"exact"`. Any playbook that currently specifies `label_mode: exact` will now behave as `one_of` without any error, potentially firing on unintended labels. If `exact` is intentionally being removed, that should be documented and callers migrated; if it is still supported it needs an explicit branch.

**2. `required: true` still set on `config.labels` (`config-schemas.ts`)**
The field is now a `tags` (array) type but remains `required: true`. There is no indication the form/validation layer handles array-empty vs. string-empty equivalence. If the tags array is `[]` but the field is marked required, validation may silently pass (empty array is truthy in JS) or always fail depending on the implementation, leading to broken playbook saves.

---

### ⚠️ Warning

**3. `_labels_match` returns `not strict` when `configured_labels` is empty and `label` is also empty (`webhooks.py`)**
This matches the old behaviour, but the new code path only reaches the fallback when `configured_labels` is falsy. If a config has both `labels: []` and no `label` key, `not strict` is returned — effectively "match everything" in non-strict mode. This may be intentional, but the comment says "legacy / fallback", which could mislead future contributors into thinking this branch is only hit for old-style configs.

**4. `autopilot-approved.yaml`, `autopilot-quick.yaml`, `autopilot.yaml` all include `ai_ready` in their trigger labels**
`ai_ready` was previously the sole trigger for `ai_ready.yaml`. Adding it to all three autopilot playbooks means that applying the `ai_ready` label will now fire **four** separate playbooks concurrently. This cross-playbook label overlap could cause duplicate work, race conditions on issue comments, or conflicting state mutations. Confirm this is deliberate.

**5. `TagInput` has no maximum tag count or length validation (`BlockEditor.tsx`)**
A user could add hundreds of tags or very long label strings. There is no `maxLength` on the `<input>` and no cap on the `value` array size. For labels that are sent to GitHub, unexpectedly long or numerous values could produce API errors at runtime rather than at configuration time.

---

### 💡 Suggestion

**6. `onBlur` auto-commit may surprise users (`BlockEditor.tsx`)**
```tsx
onBlur={() => { if (inputVal.trim()) addTag(inputVal) }}
```
Auto-committing on blur means that clicking anywhere outside the input (e.g., clicking a suggestion chip) will silently add the current partial text as a tag. Consider whether this UX is intentional, or if `Enter`/`,` should remain the only commit mechanism.

**7. String-to-array coercion in `FieldInput` for `tags` type silently drops whitespace-only tokens (`BlockEditor.tsx`)**
```tsx
value.split(",").map(s => s.trim()).filter(Boolean)
```
This is defensive and generally fine, but if the stored YAML value later round-trips through this branch (e.g. on a re-render), label order may not be preserved and comma-containing label names would be silently split. A comment noting that label names must not contain commas would prevent future bugs.

**8. `label_mode` is not exposed as a config field in `config-schemas.ts`**
The new `all_of` mode is supported in the backend but there is no UI field to set `label_mode`. Users who want `all_of` semantics must hand-edit YAML. Either add a `select` field for `label_mode` or add a hint to the `Labels` field documenting this.

---

**Summary:** 2 critical, 3 warnings, 3 suggestions — please address the `exact` mode regression and the `required` validation gap before merging.

sseshachala (commented):
## 🔍 Automated Code Review — PR #127

**fix: multi-label trigger support — closes #121 and #114**

---

### ✅ Summary

This PR refactors `_labels_match` in `webhooks.py` to support a `labels` list (plural) with `one_of` / `all_of` modes, adds a legacy fallback to the singular `label` field, and updates playbooks and the frontend `BlockEditor` accordingly. The changes are focused and well-scoped.

---

### 🔴 Critical

No critical issues found.

---

### 🟡 Warning

**1. Silent mode coercion in `_labels_match` (`webhooks.py`)**
When `label_mode` is set to an unrecognised value (e.g. a typo like `"one-of"`), the new code silently falls back to `"one_of"` instead of the previous behaviour of failing closed (`return not strict`). This means misconfigured playbooks could fire unexpectedly in strict mode.
```python
# Current (line ~404):
effective_mode = mode if mode in ("one_of", "all_of") else "one_of"
# Consider: raise a warning / log the unknown mode, or keep the fail-closed path for unknown modes in strict mode.
```

**2. `autopilot-approved.yaml` and `autopilot-quick.yaml` now share the same label list including `ai_ready`**
Both playbooks now accept `ai_ready` as a trigger label. If `ai_ready` is added to an issue and both playbooks are active, both will fire simultaneously. This may be intentional but could cause duplicate workflow executions and race conditions if they share downstream resources.

---

### 🔵 Suggestion

**1. Missing `label_mode` field in the UI schema (`config-schemas.ts`)**
The new `config.labels` (tags) field lets users specify multiple labels, but the `config.label_mode` field (`one_of` / `all_of`) is not exposed in `BLOCK_CONFIG_SCHEMAS`. Without it, users configuring triggers via the UI cannot change the matching mode — it will always silently default to `one_of`.
Consider adding a `select` field for `config.label_mode` alongside `config.labels`.

**2. `ai_ready.yaml` label change from `ai-ready` → `ai_ready` (underscore)**
The playbook now accepts both `ai_ready` (underscore, new) and `ai-ready` (hyphen, original) via `one_of`. This is good for compatibility, but existing GitHub labels likely use the hyphen form. Ensure repository labels are reconciled to avoid confusion over which label name is canonical.

**3. `TagInput` component: no `aria-label` / accessibility attributes**
The new `TagInput` React component lacks `aria-label` on the inner `<input>` and `role` attributes on the tag pills/buttons. This will fail basic accessibility audits.

**4. `onBlur` auto-commit in `TagInput` may surprise users**
The `onBlur` handler calls `addTag(inputVal)` automatically if there is any input. If a user clicks away accidentally mid-type, a partial/unintended tag will be added silently. Consider requiring an explicit commit gesture (Enter / comma) or showing a confirmation.

---

### ✅ Positives
- Excellent backward-compatibility path: the legacy `label` (singular) fallback prevents breaking existing playbooks.
- The `all_of` guard (`if incoming_label not in configured_labels: return False`) correctly prevents spurious runs when unrelated labels are added — this was the root cause of issue #114.
- The `TagInput` suggestion pills are a nice UX improvement.
- Playbook YAML changes are consistent across all four affected files.

---

**Total findings: 6 (0 critical, 2 warnings, 4 suggestions)**

PR #130, opened 2026-05-23
fix: skip fix-issue creation if one already exists for the PR

sseshachala:
## Problem
PR Reviewer creates a new \"autopilot ready\" issue on every run. If a PR is reviewed multiple times (e.g. new commits pushed), duplicate issues pile up.

## Fix
Added two blocks before `create_fix_issue` in both `pr-reviewer.yaml` and `copilot-reviewer.yaml`:
1. `check_existing_issue` — searches GitHub for an open issue with \"PR #N\" in the title
2. `issue_exists_gate` — logic block; if found, skips creation and goes straight to notify/gate_merge

## Test plan
- [ ] First review on a PR with critical issues → issue created as normal
- [ ] Second review on the same PR → no new issue created, flow continues to notify
- [ ] Different PR → new issue created independently

### DIFF
--- a/apps/api/playbooks/copilot-reviewer.yaml
+++ b/apps/api/playbooks/copilot-reviewer.yaml
@@ -105,19 +105,20 @@ blocks:
     mode: agentic
     label: Check for existing fix issue
     description: |
-      Before creating a fix issue, check if one already exists for this PR to avoid duplicates.
+      Check if a fix issue already exists for this PR.
 
       Repo: {{_trigger.repository.full_name}}
       PR number: {{_trigger.number}}
 
       Steps:
-      1. Search GitHub for existing open issues:
+      1. Retrieve the GitHub token from the Delegator credentials vault.
+      2. Search GitHub for existing open issues:
          GET https://api.github.com/search/issues?q=repo:{{_trigger.repository.full_name}}+is:issue+is:open+"PR+%23{{_trigger.number}}"+in:title
          Headers: Authorization: token <github_token>
-      2. If any issue in the results has a title containing "PR #{{_trigger.number}}", an issue already exists.
+      3. If any issue has a title containing "PR #{{_trigger.number}}", return its number and URL.
 
       Output ONLY the following JSON on the last line:
-        {"exists": true, "issue_url": "<url of existing issue>"} if found
+        {"exists": true, "issue_url": "<url>", "issue_number": <number>} if found
         {"exists": false} if not found
     next: issue_exists_gate
 
@@ -126,8 +127,40 @@ blocks:
     label: Issue already exists?
     condition: "{{check_existing_issue.exists}} == true"
     next:
-      pass: gate_merge
+      pass: append_fix_issue
       fail: create_fix_issue
+
+  append_fix_issue:
+    type: brain
+    mode: agentic
+    label: Append findings to existing issue
+    description: |
+      A follow-up review found more critical issues on the same PR.
+      Append the new findings to the existing issue and re-trigger Autopilot.
+
+      Existing issue: {{check_existing_issue.issue_url}}
+      Issue number: {{check_existing_issue.issue_number}}
+      Repo: {{_trigger.repository.full_name}}
+      Review: {{review_pr.review_url}}
+      Critical issues: {{review_pr.critical}}
+
+      Steps:
+      1. Retrieve the GitHub token from the Delegator credentials vault.
+      2. Append a comment to the existing issue:
+         POST https://api.github.com/repos/{{_trigger.repository.full_name}}/issues/{{check_existing_issue.issue_number}}/comments
+         Headers: Authorization: token <github_token>, Content-Type: application/json
+         Body: {"body": "## Follow-up review findings\n\nAI PR #{{_trigger.number}} was re-reviewed and {{review_pr.critical}} additional critical issue(s) were found.\n\nReview: {{review_pr.review_url}}\n\nPlease also address these findings."}
+      3. Remove the "autopilot ready" label (so re-adding it fires the webhook):
+         DELETE https://api.github.com/repos/{{_trigger.repository.full_name}}/issues/{{check_existing_issue.issue_number}}/labels/autopilot%20ready
+         Headers: Authorization: token <github_token>
+      4. Re-add the label (this fires issues.labeled and re-triggers Autopilot):
+         POST https://api.github.com/repos/{{_trigger.repository.full_name}}/issues/{{check_existing_issue.issue_number}}/labels
+         Headers: Authorization: token <github_token>, Content-Type: application/json
+         Body: {"labels": ["autopilot ready"]}
+
+      Output ONLY the following JSON on the last line:
+        {"issue_url": "{{check_existing_issue.issue_url}}", "issue_number": {{check_existing_issue.issue_number}}}
+    next: gate_merge
 
   create_fix_issue:
     type: brain
--- a/apps/api/playbooks/copilot-reviewer.yaml
+++ b/apps/api/playbooks/copilot-reviewer.yaml
@@ -140,17 +140,25 @@ blocks:
       Repo: {{_trigger.repository.full_name}}
       Critical issues: {{review_pr.critical}}
 
+      IMPORTANT: Create the issue WITHOUT labels first, then add the label in a
+      separate API call. GitHub only fires the `issues.labeled` webhook when a label
+      is added to an existing issue — not when an issue is created with labels attached.
+
       Steps:
-      1. Create a GitHub issue via the API:
+      1. Retrieve the GitHub token from the Delegator credentials vault.
+      2. Create the issue (no labels):
          POST https://api.github.com/repos/{{_trigger.repository.full_name}}/issues
          Headers: Authorization: token <github_token>, Content-Type: application/json
          Body:
          {
            "title": "fix: address critical review findings on AI PR #{{_trigger.number}}",
-           "body": "## Context\nAI-generated PR #{{_trigger.number}} was reviewed and {{review_pr.critical}} critical issue(s) were found.\n\nReview: {{review_pr.review_url}}\n\n## Task\nRead the review comment on the PR, identify all 🔴 Critical findings, and implement fixes.",
-           "labels": ["autopilot ready"]
+           "body": "## Context\nAI-generated PR #{{_trigger.number}} was reviewed and {{review_pr.critical}} critical issue(s) were found.\n\nReview: {{review_pr.review_url}}\n\n## Task\nRead the review comment on the PR, identify all 🔴 Critical findings, and implement fixes."
          }
-         (retrieve the GitHub token from the Delegator credentials vault)
+         Save the returned issue number.
+      3. Add the label in a separate call (this fires the webhook):
+         POST https://api.github.com/repos/{{_trigger.repository.full_name}}/issues/<issue_number>/labels
+         Headers: Authorization: token <github_token>, Content-Type: application/json
+         Body: {"labels": ["autopilot ready"]}
 
       Output ONLY the following JSON on the last line:
         {"issue_url": "<created issue URL>", "issue_number": <number>}
--- a/apps/api/playbooks/copilot-reviewer.yaml
+++ b/apps/api/playbooks/copilot-reviewer.yaml
@@ -97,8 +97,37 @@ blocks:
     label: Critical issues found?
     condition: "{{review_pr.critical}} > 0"
     next:
-      pass: create_fix_issue
+      pass: check_existing_issue
       fail: gate_merge
+
+  check_existing_issue:
+    type: brain
+    mode: agentic
+    label: Check for existing fix issue
+    description: |
+      Before creating a fix issue, check if one already exists for this PR to avoid duplicates.
+
+      Repo: {{_trigger.repository.full_name}}
+      PR number: {{_trigger.number}}
+
+      Steps:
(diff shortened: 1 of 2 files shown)


sseshachala (commented):
## 🤖 Automated Code Review — PR #130

**Fix: skip fix-issue creation if one already exists for the PR**

This PR introduces a `check_existing_issue` + `issue_exists_gate` pattern into both `copilot-reviewer.yaml` and `pr-reviewer.yaml` to prevent duplicate fix issues. The intent is clear and the flow logic is correct at a high level. However, there are several issues worth addressing before merging.

---

### 🔴 Critical

**1. Hardcoded `<github_token>` placeholder in brain block description** *(both files)*
```yaml
Headers: Authorization: token <github_token>
```
The literal string `<github_token>` is passed as instructional text to the LLM. The brain block has no mechanism to substitute this at runtime — the HTTP call will fail with a 401 Unauthorized error every time. You must reference the actual credential via a workflow secret/variable (e.g. `{{secrets.GITHUB_TOKEN}}` or whatever the framework's injection syntax is), or document explicitly how the token is injected into the agentic context.

---

### ⚠️ Warning

**2. Malformed/fragile search query URL encoding** *(both files)*
```
q=repo:{{...}}+is:issue+is:open+"PR+%23{{_trigger.number}}"+in:title
```
The double-quotes around the phrase must be URL-encoded as `%22` for the GitHub Search API to treat it as an exact phrase. Without `%22`, GitHub treats it as two separate tokens. Additionally, mixing `+` (space) with `%23` (#) inside a quoted phrase is inconsistent — use consistent percent-encoding throughout.

Recommended:
```
q=repo:{{_trigger.repository.full_name}}+is:issue+is:open+%22PR+%23{{_trigger.number}}%22+in:title
```

**3. Race condition — duplicates still possible under concurrent runs** *(both files)*
Two simultaneous review runs (e.g., rapid successive pushes) can both execute `check_existing_issue`, both receive `exists: false`, and both proceed to `create_fix_issue` before either finishes. The duplicate-prevention logic only helps for sequential runs. Consider using an idempotent label (e.g., `pr-#N-fix-tracked`) applied atomically, or a GitHub-side check before issue creation.

**4. Asymmetric `pass` routing between the two files — undocumented**
- `copilot-reviewer.yaml`: `issue_exists_gate` pass → `gate_merge`
- `pr-reviewer.yaml`: `issue_exists_gate` pass → `notify_review`

This is likely intentional (different downstream flows), but it is not documented in the PR description or inline comments. A brief comment in each file explaining why the routing differs would prevent future confusion.

---

### 💡 Suggestion

**5. Duplicate block definitions — consider a shared sub-playbook**
`check_existing_issue` and `issue_exists_gate` are copy-pasted verbatim into both YAML files. Any future bug fix or improvement must be applied in two places. If the framework supports shared/reusable blocks or sub-playbooks, extracting this logic would reduce maintenance overhead.

**6. Add `output_schema` validation for the brain block output**
The gate relies on `{{check_existing_issue.exists}}`. If the LLM returns malformed JSON or extra prose before/after the JSON, the field will silently evaluate to falsy and the gate will always create a new issue — defeating the purpose of this fix. Adding an `output_schema` or `output_format` constraint (if supported) would make the contract explicit and safe.

**7. Add `on_error` / fallback handling for the GitHub Search API call**
If the Search API returns a rate-limit error (403/429) or a network failure, the brain block will likely default to `exists: false` and proceed to create an issue. Adding an explicit error path (e.g., skip creation on uncertain state, or alert) would improve reliability.

---

### Summary

| Severity | Count |
|---|---|
| 🔴 Critical | 1 |
| ⚠️ Warning | 3 |
| 💡 Suggestion | 3 |
| **Total** | **7** |

The fix is well-motivated and the flow structure is sound, but the `<github_token>` placeholder issue (#1) will cause runtime failures and must be resolved before merging.

sseshachala (commented):
## 🤖 Automated Code Review

### Summary
This PR adds duplicate-issue prevention logic to both `pr-reviewer.yaml` and `copilot-reviewer.yaml`. The approach is sound — inserting a `check_existing_issue` + `issue_exists_gate` before `create_fix_issue` is the right pattern. The two-step label addition (create issue, then add label) is also a legitimate fix for webhook triggering. A few concerns are noted below.

---

### 🔴 Critical

**1. Hardcoded `<github_token>` placeholder in `check_existing_issue` (both files)**
> Both new `check_existing_issue` brain blocks contain the instruction:
> `Headers: Authorization: token <github_token>`
> This is the same literal placeholder string used elsewhere in these files, but unlike `create_fix_issue`, there is **no accompanying instruction** telling the agent where to retrieve the token (e.g., *"retrieve the GitHub token from the Delegator credentials vault"*). If the brain agent doesn't already know how to resolve `<github_token>`, the search API call will fail with a 401 and `check_existing_issue.exists` will be undefined — causing the gate to skip issue creation silently every time.
> **Fix:** Add the retrieval instruction to `check_existing_issue`, mirroring the wording used in `create_fix_issue` step 1.

---

### ⚠️ Warning

**2. Gate falls through to `gate_merge`/`notify_review` on undefined output**
> `issue_exists_gate` evaluates `{{check_existing_issue.exists}} == true`. If `check_existing_issue` fails (network error, bad JSON output, etc.) and `exists` is `null`/undefined, the condition evaluates to `false`, so execution falls to `fail: create_fix_issue`. This means a transient search API error causes a **duplicate issue** to be created — the very problem this PR is trying to fix. Consider inverting the gate or adding an explicit error/fallback path.

**3. URL-encoded query may not match all title formats**
> The search query uses `"PR+%23{{_trigger.number}}"` (URL-encoded `#`). This assumes the existing issue title contains exactly `PR #<number>`. The `copilot-reviewer.yaml` issue title uses `"AI PR #{{_trigger.number}}"` while `pr-reviewer.yaml` uses `"PR #{{_trigger.number}}"`. Both contain the substring `PR #N`, so the search should match — but the instruction in step 2 says *"title containing `PR #{{_trigger.number}}`"*, which also matches unrelated issues (e.g., if a manual issue is titled `"Do not fix PR #130"`). A more targeted query (e.g., filtering by `fix:` prefix or label) would reduce false positives.

**4. Identical `check_existing_issue` description block duplicated across both files**
> The two new blocks are copy-pasted verbatim between `copilot-reviewer.yaml` and `pr-reviewer.yaml`, including the generic title substring `"PR #{{_trigger.number}}"`. The copilot variant creates issues titled `"AI PR #N"`, so the check should ideally search for `"AI PR #{{_trigger.number}}"` to be precise. As-is, both files search for the same string; this works but is slightly imprecise for the copilot flow.

---

### 💡 Suggestion

**5. `issue_exists_gate` passes to different targets in each file**
> In `copilot-reviewer.yaml`, `pass: gate_merge`; in `pr-reviewer.yaml`, `pass: notify_review`. This is intentional (matching the original divergence between the two flows), but worth a comment in the YAML to make it clear this is deliberate and not a copy-paste inconsistency.

**6. No output defined for `check_existing_issue` on the error/empty path**
> The output spec only defines two cases: `{"exists": true, ...}` or `{"exists": false}`. Consider documenting that the agent should also output `{"exists": false}` on API errors rather than emitting no JSON — to ensure the gate always has a defined value to evaluate.

---

*Review generated automatically. Total findings: 6 (1 critical, 3 warnings, 2 suggestions).*

sseshachala (commented):
## 🤖 Automated Code Review — PR #130

**Summary:** This PR adds `check_existing_issue` / `issue_exists_gate` / `append_fix_issue` blocks to both `copilot-reviewer.yaml` and `pr-reviewer.yaml` to prevent duplicate "autopilot ready" issues. The intent is correct and the overall structure is sound. A few issues deserve attention before merge.

---

### 🔴 Critical

**C1 – Race condition (TOCTOU) — both files**
`check_existing_issue` (search) and `create_fix_issue` (write) are two separate, non-atomic API calls. If two pipeline runs trigger concurrently (e.g., two rapid pushes), both can pass the "issue not found" gate before either creates the issue, and duplicates are created — the exact bug this PR aims to fix. Mitigation: use a short-lived lock (e.g., a GitHub label on the PR itself set before creation and cleared after), or enforce serialization at the workflow-trigger level.

---

### ⚠️ Warning

**W1 – GitHub Search API index lag — both files**
The `search/issues` endpoint has up to ~30 s of indexing latency. A freshly-created issue may not appear in search results yet, causing `check_existing_issue` to return `{"exists": false}` even when an issue already exists. This partially undermines the deduplication guarantee. Consider a direct `list/issues` call filtered by title as a more reliable (if slower) alternative, or supplement with a local state store.

**W2 – Asymmetric routing after `append_fix_issue`**
In `copilot-reviewer.yaml`, `append_fix_issue` routes to `gate_merge`. In `pr-reviewer.yaml`, it routes to `notify_review`. If `copilot-reviewer.yaml` also has a notify step, it is silently skipped on the append path. Confirm this divergence is intentional; if not, align both playbooks.

**W3 – Non-idempotent label delete in `append_fix_issue` — both files**
Step 3 (DELETE the `autopilot ready` label) will return 404 if the label is absent (already removed by Autopilot). The agent instructions do not say to ignore 404s, so a failure here may abort the block. Add explicit error-handling guidance: *"Ignore 404 on the DELETE step."*

---

### 💡 Suggestion

**S1 – URL encoding in search query — both files**
The search URL mixes `+` (form-encoding) with `%23` (percent-encoding). Use consistent percent-encoding throughout: replace `+` separators with `%20` and quote the phrase as `%22PR%20%23{{_trigger.number}}%22` for clarity and correctness across all HTTP clients the brain agent might use.

**S2 – `is:open` filter may miss closed fix issues — both files**
If a previous fix issue was closed after Autopilot resolved it, a new push will not detect it and will create a fresh issue — again potentially duplicating tracking. Consider searching regardless of state (`is:issue` without `is:open`) and creating a new issue only if none exist at all, or rely on a dedicated label as the canonical duplicate signal.

**S3 – Code duplication across playbooks**
`check_existing_issue` and `append_fix_issue` are near-identical in both YAML files (~60 lines each). If the platform supports shared/reusable block definitions, extracting them would reduce the risk of the two files diverging silently in the future.

---

**Findings:** 1 critical, 3 warnings, 3 suggestions.

sseshachala (commented):
## 🤖 Automated Code Review

**PR:** fix: skip fix-issue creation if one already exists for the PR
**Files reviewed:** `apps/api/playbooks/copilot-reviewer.yaml`, `apps/api/playbooks/pr-reviewer.yaml`

---

### 🔴 Critical

None identified.

---

### 🟡 Warning

1. **`append_fix_issue` diverges between the two playbooks (both files)**
   - In `pr-reviewer.yaml`, the `append_fix_issue` block's comment body reads _"PR #{{_trigger.number}} was re-reviewed…"_ and its `next` is `notify_review`.
   - In `copilot-reviewer.yaml`, the same block reads _"AI PR #{{_trigger.number}} was re-reviewed…"_ and its `next` is `gate_merge`.
   - The inconsistency is intentional for the different flows, but the PR description does not document this difference. If the comment wording is ever changed it is easy to update only one file. Consider extracting the shared steps into a shared template or at minimum adding an inline comment explaining the intentional divergence.

2. **Label-cycle race condition in `append_fix_issue` (both files)**
   - The strategy of DELETE → POST label to re-fire the `issues.labeled` webhook works, but there is a window between the DELETE and the POST where another agent or webhook handler could act on the unlabelled issue. A brief race condition could result in the Autopilot never being re-triggered if the DELETE succeeds and the POST fails. Consider adding error handling / retry guidance in the description, or note this limitation explicitly.

3. **GitHub Search API rate-limit not handled in `check_existing_issue` (both files)**
   - The `check_existing_issue` brain block calls `GET /search/issues`. The GitHub Search API has a much lower rate limit (30 req/min for authenticated users vs 5000/min for core API). On busy repos reviewed many times per minute this could start returning 403/429. There is no retry or fallback instruction in the description.

---

### 🔵 Suggestion

1. **Exact-match title check is fragile**
   - The search query uses `"PR #{{_trigger.number}}"` as a substring. If the title format ever changes (e.g., `AI PR #…`), the gate won't detect existing issues. Consider pinning a unique tag (e.g., `[dup-guard:{{_trigger.number}}]`) in the issue body and searching on that to make the check more robust.

2. **`append_fix_issue` output JSON uses literal Mustache expressions**
   - The output template `{"issue_url": "{{check_existing_issue.issue_url}}", "issue_number": {{check_existing_issue.issue_number}}}` is supposed to be resolved by the brain at runtime. This is consistent with the rest of the codebase, but it's worth confirming the runtime correctly resolves nested block references in output templates — this is a slightly unusual pattern compared to the `create_fix_issue` block which emits real values.

3. **Duplicate `check_existing_issue` / `append_fix_issue` blocks in both files**
   - The two new blocks (`check_existing_issue` and `issue_exists_gate`) are nearly identical in both YAML files. If the logic needs to change in the future it must be updated in two places. Consider factoring shared logic into a reusable playbook/sub-flow if the framework supports it.

4. **Clarify `append_fix_issue` next destination in `pr-reviewer.yaml`**
   - After appending to an existing issue the flow goes to `notify_review`. This is correct for `pr-reviewer.yaml` but it is not obvious from the block description. A short inline comment in the YAML would help future maintainers understand why the flow skips `create_fix_issue` and jumps straight to `notify_review`.

---

### ✅ Summary

The fix is logically sound and addresses the duplicate-issue problem described in the PR. The two-step label approach (create without label, then add label) to correctly fire the `issues.labeled` webhook is a good catch. No critical issues found. The warnings above are worth addressing before the next iteration of this playbook.

PR #137, opened 2026-05-23
fix: API security hardening — CORS, remove unused webhooks, workspace auth

sseshachala:
### 1. CORS locked to explicit origins
- Added `allowed_origins` setting to `config.py` (comma-separated, defaults to `*` in dev)
- `main.py` reads `ALLOWED_ORIGINS` env var — set to `https://conductai.ai` on Render

### 2. Removed unused webhook endpoints
- `DELETE POST /webhooks/railway` — not in use, no signature verification
- `DELETE POST /webhooks/deploy-delegator` — leftover from Railway era, completely open

### 3. Workspace ownership checks on 3 unprotected routes
- `POST /{workflow_id}/blocks/{block_id}/compile/stream` — now verifies `workspace_id` ownership
- `GET /{workflow_id}/estimate` — now verifies `workspace_id` ownership  
- `PUT /{workflow_id}/source` — now verifies `workspace_id` ownership

Previously, anyone who knew a workflow UUID could call these endpoints.

## Render action required
Set env var on Render:
```
ALLOWED_ORIGINS=https://conductai.ai
```

### DIFF
--- a/apps/api/app/core/config.py
+++ b/apps/api/app/core/config.py
@@ -36,6 +36,9 @@ class Settings(BaseSettings):
     railway_backend_service_id: str = ""   # delegator-backend service ID
     railway_frontend_service_id: str = ""  # delegator-ui service ID
 
+    # CORS — comma-separated allowed origins; defaults to all in dev
+    allowed_origins: str = "*"
+
     # Admin — used to approve waitlisted users via POST /projects/admin/approve
     admin_secret: str = ""
 
--- a/apps/api/app/main.py
+++ b/apps/api/app/main.py
@@ -1,14 +1,17 @@
 from fastapi import FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
+from app.core.config import settings
 from app.routers import credentials, dashboard, environments, projects, runs, webhooks, workflows
 from app.routers.runs import workspace_runs_router
 
 app = FastAPI(title="Marshal API", version="0.1.0")
 
+_origins = [o.strip() for o in settings.allowed_origins.split(",") if o.strip()]
+
 app.add_middleware(
     CORSMiddleware,
-    allow_origins=["*"],
+    allow_origins=_origins,
     allow_credentials=False,
     allow_methods=["*"],
     allow_headers=["*"],
--- a/apps/api/app/routers/webhooks.py
+++ b/apps/api/app/routers/webhooks.py
@@ -3,8 +3,8 @@ Webhook endpoints for external services.
 
 POST /webhooks/slack/interactions  — Slack approval button clicks
 POST /webhooks/vercel              — Vercel deployment events (deployment.succeeded etc.)
-POST /webhooks/railway             — Railway deployment events
 POST /webhooks/github              — GitHub issue/PR events (issues.labeled, etc.)
+POST /webhooks/inbound/{id}        — Generic inbound webhook trigger
 """
 import hashlib
 import hmac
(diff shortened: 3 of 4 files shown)


sseshachala (commented):
## Automated Code Review — PR #137: API Security Hardening

This PR addresses CORS lockdown, removal of unused/insecure webhook endpoints, and workspace ownership checks on three previously unprotected routes. Overall the direction is correct and the changes are meaningful security improvements.

---

### 🔴 Critical

**1. CORS wildcard default leaves production exposed if env var is unset**
*File: `apps/api/app/core/config.py` — `allowed_origins: str = "*"`*
The default value `"*"` means that if `ALLOWED_ORIGINS` is not configured on a deployment (e.g., a staging environment or a missed Render environment variable), CORS will silently fall back to allowing every origin. This directly undermines the stated goal of the PR. The default should be an empty string (or `None`) so that a missing config causes a visible failure rather than a silent open policy. At minimum add a validator that raises a `ValueError` when the value is `"*"` and the app is not running in a recognised dev/test mode.

---

### ⚠️ Warning

**2. `allow_credentials=False` conflicts with cookie-based auth when origins are restricted**
*File: `apps/api/app/main.py`*
The middleware now restricts origins but keeps `allow_credentials=False`. If any current or future client sends cookies or `Authorization` headers as credentials, the browser will still block the request even with the correct origin. Conversely, if `allow_credentials` is set to `True` later (a common next step after restricting origins), FastAPI/Starlette will reject it unless `allow_origins` contains no wildcards — which is now satisfied. Document the intended auth mechanism to prevent a future misconfiguration here.

**3. `/webhooks/inbound/{id}` referenced in the module docstring but no implementation is visible in the diff**
*File: `apps/api/app/routers/webhooks.py` — docstring line 6*
The module-level docstring now lists `POST /webhooks/inbound/{id}` as a route, but the diff contains no corresponding `@router.post("/inbound/{id}")` handler. Either the implementation is in a separate commit (not included in this PR) or the docstring is premature. A dangling docstring entry creates misleading API documentation and should be removed until the route is implemented.

**4. Missing workspace ownership check on `estimate_workflow_cost` — `workspace_id` added but not filtered before the early-return guard**
*File: `apps/api/app/routers/workflows.py` — `estimate_workflow_cost`*
The `workspace_id` dependency is correctly added and the query now filters on it. However, the subsequent check `if not workflow.current_version_id` (early return) operates on the workspace-scoped result — this is fine. Verify that no other route in `workflows.py` that accepts a `workflow_id` path parameter is still missing the workspace filter; the diff only shows three routes patched, but a broader audit is recommended.

---

### 💡 Suggestion

**5. Split comma-separated origins parsing into a `validator` in `config.py`**
*File: `apps/api/app/core/config.py` / `apps/api/app/main.py`*
The string-split logic (`[o.strip() for o in settings.allowed_origins.split(",") if o.strip()]`) lives in `main.py` at module load time. Moving it into a `@validator("allowed_origins")` (Pydantic v1) or `@field_validator` (Pydantic v2) that returns `List[str]` would make the config self-documenting, testable in isolation, and would allow the wildcard guard from finding #1 to be enforced in one place.

**6. Trailing blank line at end of `webhooks.py`**
*File: `apps/api/app/routers/webhooks.py` — last hunk*
The diff ends with an extra blank line after the deletion of `deploy-delegator`. Minor style issue; linters like `flake8` (W391) will flag it.

---

### ✅ Positives

- Removing `POST /webhooks/railway` and `POST /webhooks/deploy-delegator` is the right call — both lacked signature verification and represented unprotected trigger surfaces.
- Adding `Workflow.workspace_id == workspace_id` to the three patched route queries is the correct pattern and follows the existing codebase convention.
- Importing `settings` at the top of `main.py` (instead of re-reading `os.environ` inline) is clean and consistent.

---

**Summary:** 1 critical, 3 warnings, 2 suggestions — 6 findings total.

sseshachala (commented):
## 🤖 Automated Code Review — PR #137
**fix: API security hardening — CORS, remove unused webhooks, workspace auth**

---

### 🔴 Critical Issues (1)

**`apps/api/app/core/config.py` — Default CORS origin is `"*"`**
The new `allowed_origins` setting defaults to `"*"`, meaning if the environment variable is not set in production, the API will allow requests from _any_ origin. This is the very misconfiguration this PR is intended to prevent. A wildcard default in a `Settings` model can silently end up in production.
```python
# Current (risky default)
allowed_origins: str = "*"
```
**Recommendation:** Either raise a `ValueError` if the field is empty in non-dev environments, or default to an empty string/`""` and require explicit configuration. At minimum, document clearly that this **must** be overridden in production.

---

### ⚠️ Warnings (2)

**`apps/api/app/main.py` — Empty `_origins` list falls through silently**
If `settings.allowed_origins` evaluates to an empty string (e.g. misconfigured env var), `_origins` will be an empty list `[]`. FastAPI's `CORSMiddleware` with `allow_origins=[]` rejects all cross-origin requests without error, which may cause silent frontend breakage rather than a clear configuration error. A startup check or log warning would improve observability.

---

**`apps/api/app/routers/webhooks.py` — `deploy-delegator` endpoint removed without deprecation notice**
The `/webhooks/deploy-delegator` endpoint is deleted entirely. If any external system (CI pipeline, Railway, manual script) was calling this endpoint, it will now receive `404` with no warning. There is no deprecation period, migration guide, or changelog entry visible in this diff.
**Recommendation:** Confirm no active callers exist before removal, or keep a stub that returns `410 Gone` with a message for a grace period.

---

### 💡 Suggestions (2)

**`apps/api/app/routers/workflows.py` — Workspace auth added inconsistently**
`workspace_id` auth via `get_workspace_id` is added to `stream_block_compile`, `estimate_workflow_cost`, and `update_workflow_source` — good. However, a full audit of all workflow endpoints should confirm _every_ DB query scoped to `workflow_id` also filters by `workspace_id`. Other endpoints in the same file may still be vulnerable to IDOR (Insecure Direct Object Reference) attacks.
**Recommendation:** Audit all `db.query(Workflow).filter(Workflow.id == workflow_id)` calls across the codebase to ensure `workspace_id` filtering is applied uniformly.

---

**`apps/api/app/routers/webhooks.py` — Railway webhook removed but config fields remain**
`config.py` still retains `railway_backend_service_id` and `railway_frontend_service_id` fields (visible in the diff context). If the Railway webhook integration is fully retired, these stale config fields add confusion. Clean them up if no longer needed.

---

### ✅ Positives
- CORS origins now properly driven by configuration rather than hardcoded `"*"`.
- Workspace-scoped authorization (`workspace_id` filtering) correctly added to three previously unguarded endpoints — good IDOR remediation.
- Removal of unused `/deploy-delegator` and `/railway` webhook endpoints reduces attack surface.

---
*Reviewed by automated code review bot. Total findings: **5** (1 critical, 2 warnings, 2 suggestions).*

PR #293, opened 2026-05-27
feat(cli): environments, credentials, and set commands (#245)

### LINKED ISSUE #245: conduct CLI: environments, credentials, and environment variables parity with UI
## Goal
Everything visible in Settings → Environments on the UI should be accessible from the CLI.

### Environments
```bash
conduct environments                          # list all environments in workspace
conduct create environment <name>             # create a new environment
conduct delete environment <name> [--yes]     # delete environment
```

### Credentials (per environment)
```bash
conduct credentials --environment <name>      # list credentials in an environment (keys only, values masked)
conduct set credential --environment <name> --key GITHUB_TOKEN --value ghp_xxx
conduct delete credential --environment <name> --key GITHUB_TOKEN [--yes]
```

### Assign environment to agent
```bash
conduct set environment --agent "PR Reviewer" --environment Production
```

## API endpoints already exist
- `GET /environments` — list
- `POST /environments` — create
- `DELETE /environments/{id}` — delete
- `GET /environments/{id}/credentials` — list credentials
- `POST /environments/{id}/credentials` — set credential
- `DELETE /environments/{id}/credentials/{key}` — delete credential
- `PATCH /workflows/{id}/environment` — assign environment to agent

## UI parity checklist
- [ ] List environments with name, credential count, assigned agent count
- [ ] Create / delete environment
- [ ] List credentials (key names + masked values)
- [ ] Set / delete credential
- [ ] Assign environment to an agent

## Why
Currently users must open the UI to configure credentials before agents can run. 
Full CLI parity means `conduct install-all` can be followed by credential setup 
in the same terminal session — no browser needed for the full onboarding flow.

sseshachala:
## Summary

- `conduct environments` — list all environments in the workspace
- `conduct create environment <name>` — create a new environment
- `conduct delete environment <name> [--yes]` — delete an environment
- `conduct credentials --environment <name>` — list credential keys (values masked) using the env-var API
- `conduct set credential --environment <name> --key GITHUB_TOKEN --value <v>` — upsert a credential
- `conduct delete credential --environment <name> --key GITHUB_TOKEN [--yes]` — remove a credential
- `conduct set environment --agent "PR Reviewer" --environment Production` — assign env to agent

## Implementation notes

- **API alignment**: uses the actual endpoints that exist — `GET/POST/DELETE /environments`, `GET/PUT /credentials/env-vars/{env_id}`, `GET /credentials/by-environment/{env_id}`, `PATCH /workflows/{id}/environment`
- **Backward compat**: `conduct create <name>` and `conduct delete <name>` still create/delete projects as before; `environment` or `credential` as first positional triggers the new paths
- **`_ENV_VAR_MAP` bridge**: `set credential` uses the flat env-var API so keys like `GITHUB_TOKEN` are automatically routed to the right `(handle, field)` pair in the database
- **No values in listing**: `conduct credentials` masks values (`ghp_***`) — no plaintext exposure

## Test plan

- [ ] `conduct environments` lists environments
- [ ] `conduct create environment Staging` creates an environment
- [ ] `conduct delete environment Staging` prompts and deletes
- [ ] `conduct credentials --environment Production` lists keys with masked values
- [ ] `conduct set credential --environment Production --key GITHUB_TOKEN --value ghp_xxx` stores correctly
- [ ] `conduct delete credential --environment Production --key GITHUB_TOKEN` removes key
- [ ] `conduct set environment --agent "PR Reviewer" --environment Production` assigns environment
- [ ] `conduct create MyProject` still works (backward compat)
- [ ] `conduct delete MyProject` still works (backward compat)

Closes #245

### DIFF
--- a/apps/web/package-lock.json
+++ b/apps/web/package-lock.json
@@ -21,7 +21,7 @@
         "class-variance-authority": "^0.7.0",
         "clsx": "^2.1.1",
         "lucide-react": "^0.460.0",
-        "next": "14.2.35",
+        "next": "15.5.18",
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
         "tailwind-merge": "^2.5.4",
@@ -33,7 +33,7 @@
         "@types/react-dom": "^18",
         "autoprefixer": "^10.4.20",
         "eslint": "^8",
-        "eslint-config-next": "14.2.35",
+        "eslint-config-next": "15.5.18",
         "postcss": "^8.4.47",
         "tailwindcss": "^3.4.14",
         "typescript": "^5"
@@ -186,7 +186,6 @@
       "version": "1.10.0",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
       "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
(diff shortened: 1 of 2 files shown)


sseshachala (commented):
## Code Review — `feat(cli): environments, credentials, and set commands (#245)`

Overall this is a solid, well-structured feature addition. The implementation follows the existing CLI patterns consistently and the UX is thoughtful (value masking, confirmation prompts, backward compat). A few issues to address before merging:

---

### 🔴 Critical

**1. Credential value exposed in process arguments (security)**
`cmd_set credential` accepts `--value <token>` as a CLI argument. On Linux/macOS, process arguments are visible to other local users via `ps aux` or `/proc/<pid>/cmdline`. A long-lived secret like a GitHub token would be transiently exposed.
**Suggestion:** Add an `--value-stdin` flag (or read from an env var) as an alternative input path so secrets don't have to pass through argv.
```python
# e.g.
if args.value == "-":
    value = sys.stdin.readline().rstrip("
")
else:
    value = args.value
```

**2. Read-then-write race condition on credential upsert/delete**
`_do_set_credential` and `_do_delete_credential` both do a GET → mutate → PUT without any locking or ETag/version check. Two concurrent CLI invocations could silently overwrite each other's changes.
**Suggestion:** Add an optimistic concurrency note in the docstring at minimum; ideally the backend API should support atomic upsert/delete endpoints instead of full-replace.

---

### ⚠️ Warning

**3. `_list_environments` ignores `workspace_id` parameter**
```python
def _list_environments(server: str, workspace_id: str, hdrs: dict) -> list:
    return api.req("GET", f"{server}/environments", hdrs)
```
`workspace_id` is accepted but never used. If the `/environments` endpoint is workspace-scoped (likely), the URL should include it (e.g. `{server}/workspaces/{workspace_id}/environments`). If the header already scopes the request, the parameter should be removed to avoid confusion.

**4. No error handling on API responses in new functions**
The existing code presumably handles API errors in `api.req()`, but the new `cmd_set` / `cmd_create` / `cmd_delete` additions do not check whether the API response contains the expected keys (e.g., `result['name']`, `result['id']`). If the API returns an error dict instead of the expected object, these will raise an unhelpful `KeyError`.
**Suggestion:** Wrap response key access with defensive checks, consistent with how errors are handled elsewhere in the file.

**5. `cmd_set` with unknown `set_command` calls `sys.exit(1)` with a vague message**
```python
else:
    print(f"Usage: conduct set [credential|environment] ...")
    sys.exit(1)
```
`set_p` is referenced here but defined only in the outer `main()` scope — this works at runtime but is fragile. Prefer `set_p.print_help()` (already done for the missing subcommand case) or at least print a cleaner error.

---

### 💡 Suggestion

**6. Masking logic is inconsistent**
In `cmd_credentials`: `masked = val[:4] + "***" if val and len(val) > 4 else "***"` — shows first 4 chars.
In `_do_set_credential`: `masked = value[:4] + "***" if len(value) > 4 else "***"` — no `None` guard (but `value` comes from argparse so it's safe here). Consider extracting a `_mask_secret(val)` helper to keep this consistent and easy to change globally.

**7. `conduct delete credential` subcommand UX is inconsistent**
`delete credential` is handled via positional `delete_args` + `--environment` + `--key` flags on the `delete` parser, while `set credential` has its own proper subparser. This is slightly inconsistent and could confuse users. Consider giving `delete` a proper subparser like `set` has, or at least document this in help text.

**8. Next.js version bump (14.2.35 → 15.5.18) in `package-lock.json`**
This is a major version upgrade (v14 → v15). Next.js 15 introduced breaking changes (e.g., async request APIs, caching defaults). The PR description doesn't mention this change or any associated migration work. Please confirm this is intentional and that the app has been tested against Next.js 15.

---

**Summary:** 2 critical, 4 warnings, 2 suggestions. The credential-in-argv security issue and the race condition on credential writes should be addressed before merging. The unused `workspace_id` parameter also needs clarification.

sseshachala (commented):
## Code Review — `feat(cli): environments, credentials, and set commands (#245)`

This PR adds environment management, credential management, and `set` commands to the `conduct` CLI. The Next.js version bump (`14.2.35` → `15.5.18`) in `package-lock.json` is a large but mechanical lockfile change. Below are findings on the meaningful logic in `main.py`.

---

### ⚠️ WARNING (3)

**1. `_list_environments` — unused `workspace_id` parameter**
`workspace_id` is accepted as an argument but never used inside the function body (the GET endpoint is `/environments`, not `/workspaces/{workspace_id}/environments`). If the parameter is genuinely not needed, remove it from the signature to avoid confusion and keep the call-sites clean.
```python
def _list_environments(server: str, workspace_id: str, hdrs: dict) -> list:
    return api.req("GET", f"{server}/environments", hdrs)  # workspace_id unused
```

**2. `_do_set_credential` — credential value length check edge case**
```python
masked = value[:4] + "***" if len(value) > 4 else "***"
```
This condition uses `len(value) > 4` (no `val and` guard), but `value` comes directly from CLI `--value` arg. If the user passes an empty string (`--value ""`), `len("")` is 0, so it correctly falls to `"***"` — fine. However, `cmd_credentials` has the guard (`if val and len(val) > 4`) which is slightly inconsistent. Unify the pattern for readability.

**3. `cmd_delete` — `credential` branch with `delete_args` is misleading**
`conduct delete credential --environment <name> --key <KEY>` dispatches to `_do_delete_credential` via `parts[0] == "credential"`, but the word `"credential"` is still consumed as part of `delete_args` positional (it shows up in nargs="+"), leading to a redundant positional. This is functional, but the UX is inconsistent with the `set credential` sub-command pattern. Consider using a proper sub-command under `delete` for consistency.

---

### 💡 SUGGESTION (3)

**4. `_resolve_environment` — N+1 fetch pattern**
Every command that calls `_resolve_environment` first fetches all environments then filters in Python. For CLIs with small environment counts this is fine, but if the API later supports filtering by name, a direct lookup would be more efficient and faster.

**5. `cmd_set` — `else` branch exits with code 1 but error may never trigger**
The `else: sys.exit(1)` in `cmd_set` can never be reached in practice because `set_command` is validated by argparse sub-parsers. The `not args.set_command` check in the dispatch section (`main()`) already handles the no-subcommand case. The `else` branch in `cmd_set` is dead code — remove it or add a comment.

**6. `package-lock.json` — Next.js major version bump without `package.json` update**
The `package-lock.json` bumps `next` from `14.2.35` to `15.5.18` (a major version), but the diff does not include a corresponding change to `package.json`. This usually indicates `package.json` was already updated in a prior commit or the lockfile was regenerated independently. Verify that `package.json` reflects `"next": "15.x"` to avoid lock/manifest mismatch warnings during `npm ci`.

---

### ✅ What looks good

- Credential values are properly masked in all display paths (`val[:4] + "***"`).
- Destructive operations (`delete environment`, `delete project`, `delete credential`) all prompt for confirmation unless `--yes` is passed — good UX safety net.
- `_do_set_credential` uses a read-modify-write (PUT) pattern that preserves existing credentials when adding or updating a single key — correct behaviour.
- Backward compatibility for `conduct create <name>` and `conduct delete <name>` (without the type keyword) is maintained.
- Argparse sub-parsers for `set credential` / `set environment` are well-structured with `required=True` on all relevant flags.

sseshachala (commented):
## Code Review — `feat(cli): environments, credentials, and set commands (#245)`

This PR adds environment management, credential management, and `set` commands to the `conduct` CLI. The Next.js lockfile bump (`14.2.35` → `15.5.18`) is mechanical. Below are findings on the logic in `main.py`.

---

### ⚠️ WARNING (3)

**1. `_list_environments` — unused `workspace_id` parameter**
The function signature accepts `workspace_id` but never uses it. Either remove the parameter or document why it is reserved for future use.

```python
def _list_environments(server: str, workspace_id: str, hdrs: dict) -> list:
    return api.req("GET", f"{server}/environments", hdrs)  # workspace_id unused
```

**2. Inconsistent credential masking guard**
`cmd_credentials` guards with `if val and len(val) > 4` (handles `None`/empty), while `_do_set_credential` uses bare `len(value) > 4`. The value from `--value` cannot be `None` at that point, but unifying the guard improves readability and prevents future bugs if the helper is reused elsewhere.

**3. `cmd_delete credential` — inconsistent routing vs `set credential`**
`conduct set credential` uses a proper argparse sub-command; `conduct delete credential` is routed by inspecting `delete_args[0]`. This means typos like `conduct delete credentials foo` will silently fall through to the project-delete path instead of surfacing a useful error.

---

### 💡 SUGGESTION (3)

**4. Dead `else` branch in `cmd_set`**
The `else: print(...); sys.exit(1)` in `cmd_set` is unreachable — argparse rejects unknown sub-commands and `main()` already handles the no-subcommand case. Remove it.

**5. `_resolve_environment` — full list fetch on every call**
Every credential/environment operation fetches all environments and filters in Python. Fine for now, but worth a comment noting the assumption that environment counts stay small.

**6. `package-lock.json` — Next.js major bump without visible `package.json` diff**
`next` goes `14.2.35` → `15.5.18`. If `package.json` was already updated in an earlier commit, this is fine — but please confirm `package.json` reflects `^15` so `npm ci` stays consistent.

---

### ✅ What looks good

- Credential values are masked in all display paths (`val[:4] + "***"`).
- All destructive operations require `--yes` or an explicit confirmation prompt.
- `_do_set_credential` uses read-modify-write (PUT) that preserves other credentials — correct.
- Backward compatibility for `conduct create <name>` and `conduct delete <name>` is maintained.
- Argparse sub-parsers for `set credential` / `set environment` enforce all required flags.

sseshachala:
test PR comment

sseshachala:
## Follow-up review findings

PR #293 was re-reviewed and 3 additional critical issue(s) were found.

Review: https://github.com/sseshachala/conductai/pull/293#pullrequestreview-4369012959

Please also address these findings.

---
_This comment was appended to tracking issue #320 for this PR._

sseshachala (commented):
## 🔒 Security Scan — sseshachala/conductai PR #293

> Scanned 2 changed file(s). Focused on newly added/modified lines only. `package-lock.json` skipped (lockfile, no code execution paths introduced).

### Summary
| Severity | Count |
|---|---|
| 🔴 Critical | 0 |
| 🟠 High | 0 |
| 🟡 Medium | 2 |
| 🔵 Info | 1 |

### Findings

**[🟡 Medium] Sensitive Data Exposure** · `packages/conduct-cli/src/conduct_cli/main.py:~389, ~401`
Credential masking reveals the first 4 characters of every secret when listing or after setting a credential:
```python
masked = val[:4] + "***" if val and len(val) > 4 else "***"
```
For many real-world secrets (e.g. GitHub PATs `ghp_…`, Stripe keys `sk_live_…`, AWS keys `AKIA…`, Slack tokens `xoxb-…`), the first 4 characters either confirm the secret *type* or — for short/low-entropy secrets — meaningfully reduce brute-force search space. Output may also be captured by terminal scrollback, screen-shares, CI logs, or `script(1)` recordings.
**Fix:** Mask the entire value (e.g. `"***" + val[-2:]` only for length confirmation, or just `"********"`). Never show a prefix of the secret. If a hint is required, show only the last 2 chars and/or the length.

---

**[🟡 Medium] Sensitive Data Exposure (CLI ergonomics)** · `packages/conduct-cli/src/conduct_cli/main.py` — `conduct set credential --value <token>`
The new `set credential` command accepts the secret as a plaintext CLI argument (`args.value`). CLI arguments are visible in `ps`/`/proc/<pid>/cmdline` to other local users, are persisted in shell history (`~/.bash_history`, `~/.zsh_history`), and frequently end up in terminal recordings and shared screenshots.
```python
def cmd_set(args):
    if args.set_command == "credential":
        ...
        _do_set_credential(server, workspace_id, api_key, token,
                           args.environment, args.key, args.value)
```
**Fix:** If `--value` is omitted or set to `-`, prompt securely via `getpass.getpass()` and/or read from stdin. Document `--value-from-file`/`--value-from-env` alternatives and discourage inline `--value` usage in `--help`.

---

**[🔵 Info] Defence in depth** · `_do_set_credential` / `_do_delete_credential`
The read-modify-write pattern (`GET` existing creds → mutate locally → `PUT` full list back) is racy: concurrent CLI invocations or a parallel UI edit will silently overwrite each other's changes, and a transient network error after the GET could lead to credential loss. Not a direct security vuln, but credential drop/clobber can have security consequences (stale creds left active, expected rotation reverted).
**Fix:** Prefer atomic per-key endpoints (`PUT /credentials/env-vars/{env_id}/{key}`, `DELETE …/{key}`) or send an `If-Match` ETag with the PUT.

---
*Powered by [Conduct AI](https://conductai.ai) · Security Scanner*

PR #399, opened 2026-05-29
Feat/eval harness

sseshachala:


sseshachala:
Closing — both changes are handled:

- **executor.py** (memory block try/catch): cherry-picked to main in da4c9b3 during the smoke test fix session.
- **RunTrace.tsx** (RunTerminalRow): superseded by this session's implementation, which has an early `return null` guard during active runs and uses `text-amber-600` for consistent styling. Merging as-is would produce a duplicate declaration and a TS error.

PR #605, opened 2026-06-12
fix: security: dev encryption key not blocked in staging/QA environme…

sseshachala:
…nts (closes sseshachala/conduct-internal#27)

### DIFF
--- a/apps/api/app/core/config.py
+++ b/apps/api/app/core/config.py
@@ -107,12 +107,11 @@ class Settings(BaseSettings):
 settings = Settings()
 
 _DEFAULT_ENCRYPTION_KEY = "dev-only-32-byte-key-change-this!"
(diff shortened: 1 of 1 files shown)


sseshachala (commented):
Small issue with the environment check — the default value for `settings.environment` is `"development"` (not `"local"`), so the current condition:

```python
if settings.environment != "local" and settings.encryption_key == _DEFAULT_ENCRYPTION_KEY:
```

...would also block local dev environments unless they explicitly set `ENVIRONMENT=local`. Breaking for anyone running locally without that env var set.

Safer allowlist:

```python
if settings.environment not in ("local", "development") and settings.encryption_key == _DEFAULT_ENCRYPTION_KEY:
```

This protects staging/QA/production while leaving local dev unaffected.

Issue #1057, opened 2026-08-10
Okta Phase 3c: E2E tests + CLI simulator + docs + staged rollout


sseshachala:
Sub-issue 3 of #1052 (Okta Phase 3 runtime enforcement). Ships everything that turns the shipped-but-flag-off code from sub-issue 2 into a real, provable, observable feature: end-to-end tests, CLI simulator, docs, and staged rollout.

**Depends on:** sub-issue 2 (bridge + config) merged first.

## Scope

**1. End-to-end integration test (`tests/integration/test_okta_e2e.py`)**

Real Okta tenant, real JWT. Requires the dev tenant to have an OAuth authorization server configured for the test service app (`0oa1677gbczxbjmcI698` — `Test-AI-Agent-RefundBot`).

Flow:

- Fetch a `client_credentials` JWT from `https://integrator-2944519.okta.com/oauth2/default/v1/token`
- Present it to a Conduct endpoint (e.g. `GET /workspaces/{ws}/agent-identities`)
- Assert it succeeds and returns the identity
- Deactivate the identity via PATCH
- Present the same JWT again → assert 401 with `Agent identity is deactivated`
- Reactivate → assert requests work again

Optional but nice: pytest fixture `okta_jwt()` that other tests can reuse.

**2. CLI simulator (`packages/conduct-cli/`)**

New command: `conduct guard simulate --as-okta-agent <jwt-or-file>`

- Takes a JWT (or reads from a file), presents it as the Bearer to `guard_check` on the configured API
- Prints what auth resolved to: identity name, source, source_id, lifecycle_state
- Explicitly labelled "simulation" so nobody uses this as a production auth path

Ships as CLI patch bump (`0.9.2` if 0.9.1 is current; adjust if bumped).

**3. Reference architecture doc (`docs/reference/okta-plus-conduct.md`)**

Phase 4 of #1036 rolled into this. Publishable — LinkedIn + blog pair.

Contents:

- Sequence diagram: builder (Claude/Gemini/etc.) → Okta AI Agent Import → Conduct sync → agent JWT → Conduct auth → Cedar policy → runtime enforcement
- Security notes: what we verify, why RS256-only, why per-workspace trust, what could still go wrong (revocation lag)
- Setup guide: how a customer configures their Okta tenant + points Conduct at it
- Comparison note: Okta identity vs Guard governance (identity plane vs authority plane)

**4. Audit log integration**

Every Okta JWT verification produces one audit event:

[...]

Issue #1052, opened 2026-08-10
Okta Phase 3: runtime enforcement — JWT verification, bridge, lifecycle gate


sseshachala:
Phase 3 of #1036. Closes the loop between "Okta identity inventory" (Phase 1, shipped) and "Guard actually gates Okta-authenticated agent actions at runtime". Without this, we have inventory but not enforcement — the identity's `lifecycle_state` and `risk_tier` don't affect real behavior yet.

## Scope

**1. JWT verification (~2 hours)**
- Fetch and cache the Okta JWKS from `https://{domain}/oauth2/default/v1/keys` (per-workspace, TTL cache)
- Verify signature, issuer, audience, and expiry on every incoming Okta-signed JWT
- Handle key rotation gracefully (retry once on unknown `kid` before failing)
- Use `PyJWT` — check if already installed transitively; if not, add explicit dep

**2. Bridge in `_resolve_agent_token`** (~30 min)
- Detect Okta JWTs by shape (three-part `xxx.yyy.zzz`, not `cond_*` prefix)
- Verify → extract `sub` (Okta client_id) → look up `agent_identities WHERE workspace_id=X AND source='okta' AND source_id=sub`
- Return the identity so downstream policy sees it as a Guard principal

**3. Enforcement** (~30 min)
- Lifecycle guard: reuse the existing pattern (deactivated/expired → 401). Already there for cond_agt/cond_api; extend behavior to Okta identities.
- Tier gating: at policy evaluation time, tier_3 identities trigger the `@advice("approval")` path. Future work — start with lifecycle only.

**4. Config (~15 min)**
- Add `issuer` and `audience` fields to the stored Okta Integration credentials (alongside existing domain + token)
- UI: one additional input on the Okta config form ("Authorization server issuer") or auto-derive from `domain`
- Per-workspace trust — never accept a JWT from an issuer not configured for the workspace

**Total: ~4 hours code + ~1-2 hours testing.**

## Testing prereqs

Cannot test end-to-end without a real Okta-issued JWT. To generate one:

1. In Okta admin (integrator-2944519.okta.com), configure an OAuth authorization server for the service app we created (`Test-AI-Agent-RefundBot`, client_id `0oa1677gbczxbjmcI698`)
2. Grant it a custom scope (e.g. `agent.act`)
3. Get a `client_credentials` token from `https://{domain}/oauth2/default/v1/token` using client_id + client_secret
[...]

PR #1062, opened 2026-08-11
fix(okta): dedupe audit events within a single request

sseshachala:
Discovered during #1057 rollout verification: hitting \`GET /auth/whoami\` with a valid Okta JWT wrote **two** identical \`auth.okta_jwt.verify\` audit rows for one request.

Root cause: \`/auth/whoami\` runs \`_resolve_okta_jwt\` twice — once via \`Depends(get_workspace_id)\`, once in the handler body to populate the response. Two calls, two audit rows.

Fix: per-request dedupe on \`(workspace_id, decision, sub)\` via a \`ContextVar\`. FastAPI runs each request in its own async context, so the var naturally resets between requests. Same token resolved N times in one request → one audit row.

New test (\`test_audit_deduped_within_same_request\`): patches \`chain_hash_for_insert\`, runs \`_resolve_okta_jwt\` twice inside a shared \`copy_context()\`, asserts \`db.add\` fires exactly once.

## Tests
\`\`\`
tests/test_okta_jwt.py             13 passed
tests/test_okta_jwt_bridge.py      12 passed (was 11)
\`\`\`

Refs #1052.


sseshachala:
Superseded by fresh branch off main — rebase collided with squash-merge history.

PR #1236, opened 2026-08-26
refactor(auth): hexagonal split of require_permission + chore(ci): make ci-py mirror

sseshachala:
## Honest scope

**Code cleanup + local CI mirror. NOT the sys.modules pollution fix I originally claimed.**

Running locally (via the new \`make ci-py\` target this PR adds) revealed the pollution has 10 sources across the test suite, not the 1 in \`test_require_permission.py\`. Fixing that one file's usage of the anti-pattern helps future maintainability but does not close the ~24 CI failures I originally claimed. Those need a systematic pass on all 10 files (tracked separately).

## What this PR actually does

1. **Splits \`require_permission\` into a pure \`check_permission\` function + thin FastAPI wrapper** (\`apps/api/app/core/auth.py\`).
   - The pure function is trivially unit-testable — no FastAPI DI, no closure state, no conftest patch to work around.
   - \`test_require_permission.py\` rewritten to import \`check_permission\` directly. No \`sys.modules\` hackery, no autouse fixtures, no \`real_require_permission\` chain. Cleaner code.
   - This is a legit architectural improvement even without the CI win.

2. **Adds \`make ci-py\` target** (\`apps/api/Makefile\`) that mirrors CI's API job exactly:
   - Isolated \`marshal_test\` database (separate from dev DB — prevents pollution)
   - Same alembic sequence CI runs (up + up/down cycle)
   - Same pytest flags CI uses (\`-q --ignore=tests/test_guard.py -m 'not matrix'\`)
   - This is what let me discover the systemic sys.modules issue in the first place. Genuinely valuable for future development.

## What this PR does NOT do

- **Does not reduce CI failure count.** Main is at ~30 failures from sys.modules pollution across 10 test files. This PR doesn't touch 9 of them. CI on main after this merges = still ~30 failures.
- **Does not fix the workspace-isolation regression** (\`test_user_a_cannot_list_workspace_b_workflows\`). That test still fails in CI because other polluting tests run before it.

## Follow-up needed (separate PR)

Apply either \`pytestmark = pytest.mark.forked\` OR the same hexagonal treatment to the remaining 9 files that mangle \`sys.modules\`:
\`\`\`
test_agent_identity_auth.py, test_analytics_dora.py, test_analytics_scorecards.py,
test_guard_mcp_auth.py, test_guard_savings.py, test_guard_spend_month_window.py,
test_guard.py, test_token_paths.py, test_workspace_seed.py, test_z_executor_lifecycle.py
\`\`\`

That's the real CI unblocker. ~30 min of mechanical work with pytest-forked.

### DIFF
--- a/apps/api/Makefile
+++ b/apps/api/Makefile
@@ -1,4 +1,4 @@
-.PHONY: schema schema-check
+.PHONY: schema schema-check ci-py ci-py-fast ci-py-help
 
 # Regenerate app/dsl/schema/v1.yaml from schema.py (source of truth)
 schema:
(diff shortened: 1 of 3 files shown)


sseshachala:
**Update:** pushed `c61cca42` — adds `make ci-py` target in `apps/api/Makefile` that mirrors CI exactly (isolated test schema, same pytest flags: `-q --ignore=tests/test_guard.py -m 'not matrix'`). Independent from the auth refactor but bundled here for delivery convenience. Full rationale in the commit body.

PR #1386, opened 2026-08-28
docs(CLAUDE): add paired-migration rule for ORM schema changes

sseshachala:
## Summary

Codifies the pattern that emerged from Sessions 1-3 of the schema-drift cleanup: every new/modified \`Column()\`, \`Index()\`, \`UniqueConstraint()\`, and \`ForeignKey()\` must ship with the paired Alembic migration in the same PR.

## What the rule says

- **Paired migration required** in the same PR for every ORM schema change under \`apps/api/app/models/\` and \`apps/api/app/modules/**/models.py\`.
- **Explicit \`name=\` required** on every \`Index()\` / \`UniqueConstraint()\` / \`ForeignKey()\`. Auto-generated names diverge between SQLAlchemy and Postgres and produce false drift.
- **Never run \`alembic revision --autogenerate\`** — proposes destructive \`DROP INDEX\` ops (see PR #1280 for the failure mode).
- **Look up existing names in \`apps/api/alembic/versions/\`** before declaring — never guess.

## Enforcement

Complementary to branch protection just enabled on main:

- \`API — lint + typecheck + test\` is a required check
- \`test_alembic_check_no_schema_drift\` runs inside that job
- Drift regressions block merge instead of accumulating

## Context

The ~68 drift ops eliminated in PRs #1337-#1374 accumulated because model changes shipped without paired migrations for months. This rule + the branch-protection gate closes that loop.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sseshachala (commented):
Looks good.

PR #123, opened 2026-05-23
Test PR #121

sseshachala:
This is a test PR for automated code review validation.

PR #1561, opened 2026-09-01
Feat/lens cli flag 1515 p2

sseshachala:
## What this PR does

One or two sentences on the change and the reason for it. Link the
issue if there is one (`fixes #123`).

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor / internal change (no user-visible behaviour change)
- [ ] Docs
- [ ] Playbook / pack

## Checklist

- [ ] Tests added or updated (or explicit note why not).
- [ ] **Bug fix only:** regression test file:line named below (a test that would fail if the bug returns). Non-negotiable for anything tagged P0 or P1. See `tests/glens/test_lens_canon.py` and `tests/guard/test_guard_canon.py` for the canon manifest — a bug fix without a regression test is unfinished.
- [ ] `pytest` (API) and `pnpm typecheck` (web) pass locally.
- [ ] No secrets, no personal data, no customer names in the diff.
- [ ] Commit messages describe the **why**.
- [ ] Docs updated if behaviour changed.

**Regression test (bug fix only):**
`path/to/test_file.py::test_name` — the test that would have caught this bug.

## How to test

Steps a reviewer can follow to see the change working.

## Screenshots or output

If UI or CLI output changed, paste before/after.

### DIFF
--- a/apps/web/src/app/(app)/lens/[sessionId]/page.tsx
+++ b/apps/web/src/app/(app)/lens/[sessionId]/page.tsx
@@ -9,6 +9,61 @@ import { API } from "@/lib/api/client"
 import { GlensDashboard } from "@/components/glens/GlensDashboard"
 import type { GlensDashboardSpec } from "@/components/glens/GlensDashboard"
 
+function SessionIdBadge({ sessionId }: { sessionId: string }) {
+  const [copied, setCopied] = useState<"" | "id" | "url">("")
+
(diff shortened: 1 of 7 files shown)


sseshachala:
Superseded by a fresh PR off main containing the one commit that didn't make it through #1560's squash-merge (75e179c2 — session ID badge + share link). All other commits on this branch were already merged via #1560. Closing to avoid mixed / stale history.

Issue #1655, opened 2026-09-06
epic(lens): migrate 40 legacy Executor tool methods into per-domain registrations


sseshachala:
## Context

`apps/api/app/modules/glens/executor.py` is 1258 LOC with 40 legacy `_tool_*` methods that predate the per-domain `ToolDef` split. Every one is registered via a shim (`_impl("method_name")` in `apps/api/app/tools/registrations/lens/_shared.py:37`) that opens a fresh DB session, constructs an Executor, and dispatches to `Executor._tool_<name>`.

New tools (`list_my_runs`, `list_runs_in_session`, all 6 actor mutators) already use the direct pattern — top-level `def impl(ctx, **kwargs)` wired as `impl=` on the `ToolDef`. This epic migrates the 40 legacy holdouts into the same shape.

## What we're changing

Per tool:
1. Move method body from `executor.py` to a top-level `def foo(ctx, **kwargs): ...` in the matching `lens/<domain>.py` file
2. Change `impl=_impl("foo")` → `impl=foo` in the `ToolDef` (one line)
3. Delete `_tool_foo` from `executor.py`

After all 40 migrate:
- Delete `_impl` + `_run` from `_shared.py`
- Shrink `class Executor` — currently only `self.agent_identity_id` is set (never read inside tools), so Executor may be deletable entirely
- Delete tests in `test_executor_guard_check.py` / `test_lens_registrations.py` that assert on the old shape (or rewrite mocks to target the new impls)

## Why

- **Single pattern:** unifies legacy tools with new tools (`impl=fn`) instead of parallel patterns
- **No per-call session overhead:** shim currently opens `SessionLocal()` on every call
- **Testable in isolation:** top-level `def impl(ctx, ...)` is a pure function; `Executor._tool_*(self, ...)` needs a fake `self.db`
- **Cuts `executor.py` from 1258 → ~50 LOC** (or deletes it)
- **Simplifies the dispatch layer:** one lookup instead of a shim chain

## Scope: 8 per-domain PRs

| Domain | Tools | LOC | Notes |
|---|---|---|---|
| guard_core | 11 | 256 | Move `MIN_SIMILARITY_SCORE` class const to module-level |
| governance | 5 | 235 | `_tool_get_compliance_status` has nested `_status` + `_grade` helpers — inline them |
| workflows | 4 | 125 | — |
| workspace | 8 | 117 | — |
| ops | 6 | 125 | — |
| marketplace | 3 | 45 | — |
[...]

PR #1662, opened 2026-09-06
refactor(lens): migrate governance tools (epic #1655 PR 7/9)

sseshachala:
5 tools migrated. Nested _status helper (closes over method-local vars) kept inline; _grade lifted to module level. Same pattern as prior PRs.


sseshachala:
Superseded by #1664 (consolidated PR that fixes the test failures from the split-PR approach).

PR #1686, opened 2026-09-07
fix(runtime + lens): cred cache + Slack unfurl + actor ident + awaiting-approval stats

sseshachala:
Four symptoms from the 2026-09-07 Lens → Guard-pause → Slack demo:

## What broke

| # | Symptom | Where |
|---|---|---|
| 1 | Slack card said **Requester: unknown** even though the run was triggered by \`lens:user_...\` | `guard_block.py` never passed \`requester_agent_ident\` to \`create_approval_request\` |
| 2 | Both \`/runs/...\` and \`/theguard/approvals/...\` URLs unfurled to the generic homepage OG card | \`post_message\` sent Block Kit without disabling unfurls; auth-gated URLs unfurled to nothing meaningful |
| 3 | Run detail StatRow showed **Turns — · Tokens — · Est. cost —** while paused | Reads as "no data"; should read as intent |
| 4 | Warning triangle next to Approve/Reject | Slack's "some links didn't unfurl" indicator — side-effect of #2 |

## The fix

- **guard_block.py** — fetch \`run.triggered_by\` and pass as \`requester_agent_ident\`. Slack card now shows \`Requester: lens:user_3FaV...\` instead of "unknown".
- **slack.py** — \`post_message\` auto-sets \`unfurl_links=false, unfurl_media=false\` when blocks are attached. Block Kit owns its CTA; suppressing unfurl also kills the ⚠️ indicator (#4).
- **RunDetailPanel.tsx** — when \`isAwaiting(status)\` AND no LLM activity, collapse Turns/Tokens/Est.cost into a single "Waiting for approval" cell spanning the three columns. Keeps StatRow density, reads as intent.
- **runUtils.ts** — \`formatTrigger\` unwraps \`lens:<clerk_user_id>\` → \`Lens · abcdef\` so the Triggered-by chip is human-readable.

## Scope kept small (ponytail)

- **No Clerk display-name resolver.** That needs a user cache we don't have. Raw ident kills the customer-visible "unknown" and unblocks the demo. If we later want first/last names, add a \`clerk_user_cache\` table populated on first token verify — one follow-up.
- **No new column on \`runs\`.** Resolution happens at read time (Slack card, frontend chip) via the shared helper approach; no migration, no backfill.

## Test plan

- [x] \`test_guard_approval_requester_ident.py\` — asserts \`guard_block\` propagates \`run.triggered_by\` into \`create_approval_request(requester_agent_ident=...)\`. Passes.
- [x] Frontend tsc clean.
- [ ] **Live smoke** (post-merge): trigger a Lens → run → Guard pause → Slack card. Verify:
  - Card shows \`Requester: lens:user_...\` (not "unknown")
  - No unfurl cards below the Approve/Reject buttons
  - No warning triangle next to buttons
  - Run detail shows "Waiting for approval" instead of three em-dashes
  - Triggered-by chip reads "Lens · abcdef"

## Follow-ups (out of scope)

- Clerk display-name cache → real names on Slack + frontend
- Same \`requester_agent_ident\` propagation for the DSL-level \`approval_block\` (this PR only fixes the Guard-rule path)

### DIFF
--- a/apps/api/app/modules/guard/approval.py
+++ b/apps/api/app/modules/guard/approval.py
@@ -49,6 +49,39 @@ def approval_url(request_id: str | _uuid.UUID) -> str:
     return f"{_base_url()}/theguard/approvals?highlight={request_id}"
 
 
+def _format_requester(requester_email: str | None, requester_agent_ident: str | None) -> str:
+    """Render a human-friendly requester line for the Slack card.
+
+    Prefers a resolved Clerk user (name + email) when the ident is a Lens
+    actor. Falls back to the raw ident so we never show "unknown" when we
+    at least know who kicked off the run.
+    """
+    if requester_email:
+        return requester_email
+    if not requester_agent_ident:
+        return "unknown"
+    ident = requester_agent_ident
+    if ident.startswith("lens:"):
+        clerk_id = ident.split(":", 1)[1]
+        try:
+            from app.core.auth import get_clerk_user_info
+            info = get_clerk_user_info(clerk_id)
+        except Exception:
+            info = {}
+        name = info.get("name") if isinstance(info, dict) else None
+        email = info.get("email") if isinstance(info, dict) else None
+        if name and email:
+            return f"{name} <{email}>"
+        if name:
+            return name
+        if email:
+            return email
+        # Clerk lookup failed (offline / no secret / user gone) — show a
+        # short tail rather than the full opaque ident.
+        return f"Lens · {clerk_id[-6:]}" if len(clerk_id) > 6 else ident
+    return ident
+
+
 def _snapshot_input(tool_input: dict | None) -> dict:
     """Redact secrets from the tool input before persisting. Bound size at
     ~8KB serialized — large blobs don't add signal for a human deciding."""
@@ -185,7 +218,7 @@ def dispatch_approval_notifications(
     url = approval_url(request.id)
     rule_id = request.rule_id
     summary = request.rule_message or f"Approval required for rule {rule_id}"
-    requester = request.requester_email or request.requester_agent_ident or "unknown"
+    requester = _format_requester(request.requester_email, request.requester_agent_ident)
 
     # Workflow context — when this pause came from a workflow run, look up the
     # workflow name + build a run-page link so the Slack post is actionable
--- a/apps/api/app/runtime/blocks/guard_block.py
+++ b/apps/api/app/runtime/blocks/guard_block.py
@@ -212,6 +212,25 @@ def _execute_guard(
             # emits an approval_requested event with the approval_id + URL.
             from app.runtime.exceptions import ApprovalRequired
             from app.modules.guard import approval as _approval
+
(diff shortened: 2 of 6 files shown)


sseshachala:
Wired `get_clerk_user_info` (LRU-cached, `apps/api/app/core/auth.py:189`) into the approval card via a small `_format_requester` helper.

Renders as (in fallback order):
- `Alice Smith <alice@example.com>` — Clerk returns both
- `Alice Smith` — name only
- `alice@example.com` — email only  
- `Lens · abcdef` — Clerk lookup failed / returned nothing
- `webhook:inbound` / `schedule:...` etc. — non-lens idents pass through unchanged
- `unknown` — only when we truly have no ident (unchanged from before)

+6 unit tests covering each branch. All 8 pass.

Original memory note ("we don't have a Clerk cache") was wrong — I missed the existing helper. Corrected in commit af67a713.

sseshachala:
## Added: in-process credential cache (kills intermittent MissingProviderKey on resume)

**Symptom** (run 440c3a68... vs 97aaf78c... on workflow a39909a0...): same workflow, block 3 (`report_outcome`) hit `MissingProviderKey` on the failed run, succeeded on the good one. Both runs paused after block 2 and resumed. Different resume behavior, same DSL.

**Root cause: two co-conspirators, one fix.**

- **Bug A:** `fetch_credential` (`apps/api/app/core/credentials.py:131`) swallowed every non-200 broker response into `{}` with zero telemetry. 401 from stale cred_token, 500 from broker restart, timeout — all indistinguishable, all silent. `brain_block`'s ladder then read `{}` as "no key configured" and raised, blaming the user's env.
- **Bug B:** on resume, the executor's finally-block invalidates the cred_token from the previous segment (`executor.py:453+`). If a block in the resumed segment hit the broker with the state-recovered stale token before a fresh one was minted, broker returned 401 → (A) swallowed → boom.

**Correct fix.** Eliminate the per-block broker round-trip for in-process callers. Executor already fetches every workspace credential once via `get_all_credentials()`. New module `apps/api/app/runtime/run_credentials.py` snapshots that map into a process-local dict keyed by `cred_token`, populated at run init, purged in the finally-block. `fetch_credential` consults the snapshot first; broker call remains only for out-of-process callers (CLI, remote workers) and for the vanishing cache-miss window.

Same executor init runs on resume → segment N gets a fresh snapshot → Bug B disappears as a side-effect of removing the failure surface.

`fetch_credential` also grows structured log lines on both broker exception and non-200 so operators can see the failure class when the broker fallback *does* fire — Bug A safety net for CLI callers.

**Tests:** `test_run_credentials_cache` — populate/resolve/purge lifecycle, cache-hit skips httpx entirely, cache-miss falls back to broker, broker exception returns `{}` without crashing, non-200 returns `{}`. 8/8 pass. Combined with the prior 8 guard-approval tests: **16/16 pass**.

**Zero API changes** to `fetch_credential` — every caller in `blocks/*.py` picks up the fix without migration.

Commit: 6e2e2756

Issue #1594, opened 2026-09-03
Guard: record-only anomaly check for tool-call argument drift, filesystem writes first


yassergb26:
Proposal for a record-only argument anomaly check under `apps/api/app/guard/` that learns per-key baselines for tool-call argument values and records an advisory finding when a call lands far outside that key's own history. The key is `(workspace_id, agent_id, tool_name, arg_path)`, agent_id being the calling identity on the guard_check path today, so a value is judged against what that agent in that workspace normally passes to that argument, not a global norm. Storage is one Postgres table of per-key rolling aggregates: Welford count, mean and variance for numeric args, capped sets of workspace-salted value hashes for string args (raw values are never stored; paths reduce to a directory-prefix-plus-extension feature); no Redis in v1, hot counters can come later if write volume demands it. "Far outside" numerically: a z-score of 6.0 or more against the key's own mean and variance for numeric args, a never-seen hash for categorical args (novelty auto-disarms once a key exceeds 200 distinct values, since that shape is free text, not a category), and nothing fires until a key has 200 recorded calls, so cold start stays silent. 6.0 and 200 are conservative starting values and the false-positive rate on real traffic is to be measured before any tightening or enforcement. First PR: filesystem write tools only (the existing filesystem-write tool group), behind a config flag defaulting off, record-only; findings land as advisory audited events in the hash-chained audit trail on the run and never touch BLOCK or APPROVAL. Naming note: this is argument drift, separate from the existing guardrail-config drift alert tier.

sseshachala:
Thanks for the depth here — clean writeup and clearly you read the code (guard_check path, filesystem-write tool group, existing config-drift tier all named correctly).

Four things I'd want pinned in this thread before writing code:

**1. Storage bound.** The key tuple `(workspace_id, agent_id, tool_name, arg_path)` can grow unbounded — if a tool takes JSON args with dynamic keys, `arg_path` cardinality explodes on its own. Can you propose a hard per-workspace row cap and an LRU eviction policy? Actual numbers matter since we're multi-tenant.

**2. Positioning fit.** Guard's public framing is *"before consequential actions execute"*. This proposal is *after-the-fact drift signaling*. Different mental model. I'm fine adding it, but I'd land it as a new module (e.g. `apps/api/app/modules/behavior/`) rather than under `apps/api/app/guard/`. Keeps the two stories clean. Does that split work for you?

[...]

PR #1715, opened 2026-09-08
behavior: record-only argument anomaly check for filesystem writes (#1594)

yassergb26:
Implements the argument anomaly check from #1594, in the shape agreed in that thread: new module, record-only behind a flag defaulting off, advisory findings in the audit trail, no enforcement.

What it does: guard_check calls for tools in the filesystem-write group are scored against per-key rolling baselines in a new behavior_arg_baselines table, keyed on (workspace_id, agent_key, tool_name, arg_path). Numeric args keep Welford count/mean/variance and flag at a z-score of 6.0 or more; string args keep capped sets of workspace-salted value hashes and flag never-seen values, with novelty auto-disarming past 200 distinct values. Nothing fires below 200 recorded calls per key. Paths reduce to a directory-prefix-plus-extension feature before hashing, dotfiles keep their own bucket so .env stays distinct from an extensionless README, and raw string values are never persisted.

Positioning, per your note: this lives in apps/api/app/modules/behavior/, not under guard/. The one thing that stayed on the Guard side is the arg_anomaly_enabled flag, which sits on guard_config and rides the existing GET/PATCH /guard/config behind require_permission("guard.settings.edit"). It gates behavior on the guard_check surface, and a separate config table plus endpoint for one boolean would be the single-call-site abstraction CONTRIBUTING warns about. Say the word if you want it moved.

Storage bound: two caps. MAX_TRACKED_ARG_PATHS (64) bounds fan-out within one (workspace, agent, tool) so a tool with dynamic JSON keys cannot dominate, and MAX_ROWS_PER_WORKSPACE (5000) is the tenant ceiling with least-recently-updated eviction on insert, as agreed. updated_at is the LRU key and there is a (workspace_id, updated_at) index for the eviction scan. Rows the current transaction already holds locked are excluded from eviction, so a row being written cannot be evicted underneath itself. Both counts are lazy: they only cost a query when a call actually creates a new key, which is rare once a baseline is established.

Scope: exactly TOOL_GROUPS["filesystem-write"] via normalize_tool, so write, edit, write_file, edit_file, str_replace_editor. The paxel-versus-raw name reconciliation is deliberately not in this PR.

Wiring: one fail-open block in guard_check_impl, placed after policy matching and running only for calls that end allowed or audited, so blocked calls and approval polling never train the baseline. Findings are recorded through the existing _record_event path as audited events (hash-chained), with a new optional rule_message argument carrying the arg_path and score for triage. The `Decision` and the string returned to the agent are untouched on every path. Migration 0113 is hand-written, creates the table and adds the flag, and pairs with the model changes in this PR.

Known v1 limits:

- The CLI hook path is unobserved, raw args never reach the server there.
- agent_key is the calling user identity until a surface resolves an agent identity.
- A flagged outlier is still folded into its baseline, so sustained drift flags its first occurrences then normalizes.
- An arg with just under 200 legitimate distinct values will flag rare-but-valid values.
- An arg constant for 200 calls flags on any change, without reference to the 6.0 threshold, since a constant has no variance to score against. Expect this to be the highest-volume numeric rule.
- File content and patch text (content, old_string, new_string, text, body, patch) are observed by length only at any size. Treating a short one as a category would mint a novel value on almost every write and swamp the measurement.

6.0 and 200 are starting values, and the false-positive rate on real traffic decides what changes at the roughly two week checkpoint you suggested.

Tests: 34 unit tests (flattening, Welford scoring and thresholds, novelty and cardinality disarm, row init before flush, fingerprint-before-storage, both caps, and LRU eviction including workspace scoping, protected rows and skip-locked). Full API suite, 1524 passed. `alembic upgrade head` runs clean from an empty database and `alembic check` reports no schema drift. Local smoke against docker compose, seeded past the 200-call minimum:

```
=== 1. flag OFF (default): 5 writes ===
response: {"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"[ws:22222222] ok"}]}}
anomaly events (expect 0): 0

=== 2. flag ON, cold start: write .env with 100000-char content ===
anomaly events (expect 0): 0        <- below the 200-call minimum, stays silent

=== 3. 220 normal writes seeded (docs/*.md, content 200-260 chars) ===
baseline rows:
write|content#len|numeric|221
write|file_path|category|221
write|mode|category|221

=== 4. the call that SHOULD trip: write to .env with 100000-char content ===
response (must still be ok, record-only):
{"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"[ws:22222222] ok"}]}}
audited|behavior.arg_anomaly.numeric|arg 'content#len' value 1e+05 is 14.8 standard deviations from its baseline (n=221)

=== 4b. a never-seen path: write to /etc/passwd ===
response: {"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"[ws:22222222] ok"}]}}
audited|behavior.arg_anomaly.novel_value|arg 'file_path' value unseen in 223 prior calls for this key (value stored as hash only)

=== 5. the call that should NOT trip: one more normal write ===
anomaly events before=1 after=1 (must be equal)

=== 6. proof nothing was blocked ===
allowed|229
audited|1

=== stored baselines hold fingerprints only ===
 mode      | category | 224 | 97eb6566b9092e2b
 file_path | category | 224 | 6211b29f9c8dcc5d
 file_path | category | 224 | 6aa98268aab12e30
 file_path | category | 224 | e54b9aa6fcf205da
```

Follow-ups, out of scope here: enforcement once the false-positive rate is known, the paxel-versus-tool_groups name reconciliation, and per-agent-identity keying if a surface starts resolving one.


sseshachala:
Welcome, and thanks for the PR — this is the repo's first outside contribution and the quality bar you set is a great signal. The write-up on scope, storage bounds, known v1 limits, and the live smoke transcript is exactly what makes a substantive PR reviewable.

**Merge outlook: yes, with one micro-ask below.** The design is sound (record-only, fail-open, capped growth, hash-only storage, 34 tests) and matches #1594.

**Answers to the questions you flagged:**

1. **Positioning (module under `behavior/` vs `guard/`).** Keep it at `apps/api/app/modules/behavior/`. Your instinct is right — the abstraction is "watch tool-call shape and score anomalies," which is a substrate any surface (Guard, Lens, workflow blocks) might want to consume later. Coupling it to `guard/` would lock it to one caller. The one `guard_config.arg_anomaly_enabled` flag riding the existing `GET/PATCH /guard/config` endpoint is the right compromise — a separate config table and endpoint for one boolean would be exactly the single-call-site abstraction CONTRIBUTING warns against. No change needed here.

2. **Migration collision.** We had an unmerged `0116_backfill_workspace_users_from_owner_id.py` in a parallel branch. Just renamed it to `0117` on our side so yours can land clean at `0116`. You do not need to rebase.

**One optional ask (nice-to-have, not a blocker):** consider promoting the two magic constants — `Z_SCORE_THRESHOLD = 6.0` and `MIN_CALLS_BEFORE_FLAG = 200` — from module constants to columns on `guard_config` (or a small `behavior_config` if you'd rather keep the boundary clean). Rationale: you called out a two-week false-positive checkpoint, and tuning those two values via a config PATCH beats a code deploy. If you'd prefer to defer this to a followup PR after real traffic tells us what the values should be, that's a fine call too — I'll file it as a separate issue if you skip it here.

Also: v1 known limits are honestly documented and match what I would have raised in review, particularly:
- CLI hook path unobserved (agreed, out of scope).
- Constant args flagging on any change once baseline holds — worth a comment at the flag site so the audit reviewer knows why they're seeing a novel-value finding on what looks like normal traffic.
- File-content observed by length only (correct call — treating short content as a category would swamp the measurement).

Thanks again — genuinely nice work.

yassergb26:
Both done, and one thing changed on the migration number.

Thresholds are now workspace-tunable. arg_anomaly_zscore_threshold and arg_anomaly_min_samples are nullable columns on guard_config, riding the same GET/PATCH /guard/config surface as the enable flag. NULL means use the module default, so the default value itself still lives in exactly one place in the code and cannot drift from the schema. Out-of-range values are rejected at the endpoint (422) and anything unparseable falls back to the default at read time rather than raising, since an advisory path should not cost a finding over a bad config row. Verified end to end: with the columns set to 2.0 and 5, a call flagged at n=8 that the defaults keep silent.

Constant-baseline comment added at the flag site, and the finding message now says so out loud: "value 3 is the first change from a constant baseline of 1 in 240 calls (no variance to score against, so the z-score threshold does not apply)".

On the migration: main moved again after your note. 0116_guard_audit_share_token_hash landed with revision 0116 and down_revision 0115, which is the same slot you had cleared for me, so I rebased onto current main and took 0117 chained onto it. alembic heads now reports a single 0117 and alembic check reports no drift from an empty database. If your renamed backfill still needs 0117, say so and I will move to 0118.

Rechecked after the rebase: 39 behavior tests pass, and I diffed the full-suite failures against clean main on the same machine to be sure I had not introduced any. Identical sets, so nothing here regressed. pytest-forked 1.7.5 cannot do fork isolation on Windows, so a full local run there shows 8 failures that pass in isolation and are unrelated to this PR.

sseshachala (approved):
Approving — well-scoped record-only v1 that matches what we agreed in #1594.

**What holds up on a close read**
- Fail-open path: observation error → `db.rollback()` → decision string untouched.
- Row growth bounded twice (per-tool fan-out + tenant LRU) with `SKIP LOCKED` so eviction never wedges a live call.
- Workspace-salted 16-hex fingerprints; raw strings never persist; free-text args (`content`, `old_string`, `new_string`, ...) reduce to `#len` only — right call to avoid swamping novelty.
- Migration 0117 hand-written, revises 0116 (clean vs current `main`), covers all three `guard_config` columns, has a proper downgrade, and uses named constraints/indexes per repo convention.
- `_new_baseline` explicitly sets stat attrs so freshly built rows are immediately scorable (column defaults apply at flush, not construction) — subtle and correct.
- 34 unit tests + a docker smoke that proves cold-start silence, 6σ trip, novel-path flag, and `allowed=229 / audited=1 / blocked=0`.

**Two follow-ups we'll take on (not blocking merge — both are pre-enforcement concerns)**
1. `observe()` calls `db.commit()` on line 458, which commits the whole session, not just the baseline writes. Fine for record-only, but before enforcement lands this should become a savepoint (`with db.begin_nested():`) or hoist to the caller.
2. `_will_allow = rule is None or _advisory or rule.get("action", "audit") == "audit"` predicts the decision instead of reading it. Cleanest fix is to move `observe()` after the decision is finalized and gate on the recorded decision string. If policy semantics ever add a new action or downgrade path, this heuristic silently pools blocked/approval-gated data into baselines.

Both live in `mcp_impls.py` — our transactional conventions, our fix. Filing a follow-up issue.

Thanks for the tight write-up and the smoke transcript — made this an easy review.

sseshachala:
@yassergb26 — separate from the review: this is the first outside contribution to Conduct, and it's a strong one. You picked a design-heavy issue, held to the shape we agreed in #1594, and shipped it with an unusual amount of care: a hand-written migration with a real downgrade, 34 unit tests that cover the not-obvious cases (cold-start silence, constant-baseline flag, cardinality auto-disarm, LRU eviction with protected rows and skip-locked), a docker smoke transcript that proves record-only end-to-end, and a PR description that anticipated most of the review questions. Reviewer effort dropped to near zero because of it.

The two follow-ups noted on the review are our code and our conventions to fix, not yours. Happy to have you back on the next thing whenever you have cycles.

Thank you.

sseshachala:
Hey @yassergb26 — took the liberty of rebasing this onto latest `main` for you so we can move it forward. `maintainerCanModify` was on, so the force-push went to your branch directly (you should see the new commit).

## What I resolved

Since you opened this on Sept 8, main shipped migrations 0117 through 0124 and a chunk of `GuardConfig` changes, so a straight rebase would have collided. Handled all mechanically — no logic changes to your code.

| Conflict | Resolution |
|---------|------------|
| `alembic/versions/0117_behavior_arg_baselines.py` | Renumbered to `0125_*`, `down_revision "0116" → "0124"`. DDL untouched. |
| `guard/models.py` `GuardConfig` columns | Kept both `inbox_auto_close_days` (from #1879) and your `arg_anomaly_*` fields. |
| `guard/routers/config.py` — `ConfigOut` / `ConfigPatch` / `_config_to_out` / PATCH handler | Merged both sides so auto-close and arg-anomaly config live alongside each other. |

Also had to run through `mcp_impls.py` after #1846 — that PR deleted the security-loop code surface (`post_finding_impl` / `trigger_fix_impl`) between the time you opened and today. Your code didn't depend on it, so nothing needed changing.

## Verified

- `pytest tests/behavior/ tests/guard/test_inbox_auto_close.py tests/guard/test_guard_inbox_router.py -q` → **51 passed / 0 failed**
- Alembic single head at `0125`, no drift
- `pnpm typecheck` clean (no web-side conflicts)

## Ready to merge from my side

The feature flag defaulting off means end-user impact is zero until an admin flips `arg_anomaly_enabled` — safe to land whenever. Ping me if the rebase looks off or you want to add anything before it goes in.

Really nice piece of work — the salted-hash / Welford split is exactly the right shape for record-only observation. Looking forward to seeing what surfaces once we can flip the flag on a real workspace.

PR #1785, opened 2026-09-11
fix(lens): approve/reject flips run status optimistically

sseshachala:
## Summary
- Approve/Reject in Lens now updates the panel **on click**, not after the network round-trip.
- Three handlers converted (RunBubble, ActionConfirmBubble, RunDetailPanel.handleApproval).
- If the POST fails, status reverts and the error surfaces via existing actionErr/runActionErr state.

## Root cause
Every approve handler awaited `authFetch(/approve)` before touching visible state. Between click and network response the buttons only dimmed — the panel looked frozen until an SSE \`run.status_changed\` event landed. Users read that as "nothing is happening."

## Fix
Flip \`run.status\` (or \`runData.status\`) locally on click. \`_postAction\` / \`_postRunAction\` gained an optional \`onError\` callback that reverts state if the request fails. SSE still confirms the transition on success — no double-flip because status is already at the target value.

## Test plan
- [ ] Run a workflow that pauses for approval; click Approve → buttons vanish immediately, panel shows running state, no visible wait
- [ ] Same, click Reject → panel shows cancelled state immediately
- [ ] Simulate a 500 from /approve → panel reverts to paused, error banner shows
- [ ] Approve from the RunDetailPanel Approvals tab → same instant behavior
- [ ] SSE stream lands with run.status_changed → no visible re-flicker (status already matches)

_Independent of #1784 (refactor). Either can merge first; the loser rebases mechanically._

### DIFF
--- a/apps/web/src/components/glens/GLensChatPage.tsx
+++ b/apps/web/src/components/glens/GLensChatPage.tsx
@@ -1049,7 +1049,7 @@ function ActionConfirmBubble({
 
   const isMutation = toolName !== "decide_approval"  // heuristic — decide is itself an approve/reject
 
-  async function _postRunAction(url: string, body?: unknown, onOk?: (data: Record<string, unknown>) => void) {
+  async function _postRunAction(url: string, body?: unknown, onOk?: (data: Record<string, unknown>) => void, onError?: () => void) {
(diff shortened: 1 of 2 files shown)


sseshachala:
Superseded by the rebased branch — reopening as a fresh PR against post-#1784 main.

Issue #47, opened 2026-05-21
feat: Memory Vector block — cross-run context and RAG over codebase


sseshachala:
## Memory Vector block

Currently in the palette but not backed by any implementation. Remove from palette for now — add back when there's a real use case.

## When to re-add
- Agent needs to remember context across multiple runs (e.g. "what did we change last time?")
- RAG over large codebase where repo doesn't fit in a single Brain block context
- Cross-run learning — agent improves over time based on past run outcomes

## What needs to be built
- Vector DB integration (Pinecone, pgvector, or similar)
- Read/write executor logic for the memory block
- UI config fields (collection name, top-k, embedding model)
- Decide scope: per-workspace or per-workflow memory

## Phase
Phase 5+ — not needed for current autopilot use cases.

Issue #46, opened 2026-05-21
feat: add PR Reviewer, Deploy & Notify, Incident Responder templates


sseshachala:
## New agent templates to add

Priority order:

### 1. PR Reviewer
- Trigger: GitHub `pull_request.opened`
- Blocks: fetch PR diff (GitHub tool) → Brain (AI reviews diff, checks style/bugs/security) → GitHub tool posts review comment back to PR
- Tags: GitHub

### 2. Deploy & Notify
- Trigger: Inbound webhook (from CI/CD pipeline)
- Blocks: Brain (parse deploy payload) → Tool (trigger Vercel or Railway deploy) → Output (Slack notify with result)
- Tags: GitHub, Slack, Vercel

### 3. Incident Responder
- Trigger: Inbound webhook (PagerDuty / OpsGenie)
- Blocks: Brain (investigate — reads recent commits, logs) → Output (Slack summary with root cause hypothesis)
- Tags: Slack

### 4. Dependency Updater
- Trigger: Run via `marshal run` on a cron schedule
- Blocks: Brain (scans for outdated deps, bumps versions) → push branch → open PR
- Tags: GitHub

## Acceptance criteria
- Each template is a YAML playbook in `apps/api/playbooks/`
- Selectable from the New Agent page
- Canvas opens pre-wired with correct blocks and edges
- Config form shows only relevant fields per block

sseshachala:
## Additional template: Autopilot Full + Approval Gate

Same as Autopilot Full but adds a human approval step before the PR is pushed.

### Flow
```
fetch_issue → implement_fix → run_tests → tests_pass (logic)
  → pass: approve_pr (approval gate) → push_pr → notify_success
  → fail: notify_failure
```

### approve_pr block
```yaml
approve_pr:
  type: approval
  label: Approve PR
  description: "Review the fix for issue #{{fetch_issue.issue_number}} before pushing to GitHub"
  next:
    approved: push_pr
    rejected: notify_failure
```

### Why
- Core to the product pitch: "AI does the work, human approves before it ships"
- Gives teams confidence to run autopilot on production repos
- Approval notification goes to Slack with a one-click approve/reject button

### Template name
`autopilot_approved` — selectable from New Agent page alongside Quick and Full

PR #1790, opened 2026-09-11
fix(sec): S01+S05+S12 — hardened trial signup surface

sseshachala:
## Summary

Hardens the trial signup surface end-to-end. Fixes three findings from the 2026-09-11 external audit at once because they all live in the same request path.

- **S01 (P0 Critical)** — `/guard/trial/provision` no longer returns any user's decrypted token. Replaced with a two-step flow: `/provision` mints a challenge + emails a verification link (generic 200 accepted regardless of whether the email exists); `/redeem` consumes the challenge (single-use, 30-min TTL) and only issues tokens for verified NEW emails. Existing emails resolve to a sign-in URL — never a token.
- **S05 (P0 High)** — `/guard/trial/ops` no longer accepts a tenant `guard.spend.view_all` permission. New `require_platform_operator()` dep gated on `PLATFORM_OPERATOR_CLERK_IDS`. Nobody has ops access by default; grant on-call staff explicitly.
- **S12 (P1 High)** — X-Forwarded-For handling no longer trusts the first (attacker-controlled) hop. Rate limits (per-IP, per-email, global) all fail CLOSED on Redis outage — no free credential issuance without abuse controls.

## What ships

| Area | Change |
|---|---|
| `apps/api/app/core/config.py` | `platform_operator_clerk_ids`, `trusted_proxy_cidrs` |
| `apps/api/app/core/auth.py` | `require_platform_operator()` dep |
| `apps/api/app/core/email.py` | `trial_verify` fallback template |
| `apps/api/app/modules/guard/trial_challenges.py` | Redis-backed challenge store + trusted-proxy IP parser (new file) |
| `apps/api/app/modules/guard/routers/trial.py` | `/provision` challenge mint, new `/redeem`, `/ops` gate flip |
| `apps/web/public/install.sh` | Two-step CLI flow (paste verify URL back after clicking email link) |
| `apps/api/tests/guard/test_trial_provision.py` | Rewritten for new shape |
| `apps/api/tests/guard/test_trial_challenges.py` | Unit tests for store + XFF parsing (new) |

## Deploy checklist

Merging this alone doesn't finish the fix — three env vars need to be set in Render **before** rollout:

1. `PLATFORM_OPERATOR_CLERK_IDS` — comma-separated Clerk user IDs of on-call staff who need `/guard/trial/ops`. Anyone not in this list gets 403.
2. `TRUSTED_PROXY_CIDRS` — the Render / Cloudflare LB CIDR(s). Without this, X-Forwarded-For is ignored and every request appears to come from the same LB IP → rate limits collapse into one shared bucket. Setting it correctly makes per-IP throttling per-real-client.
3. Verify `RESEND_API_KEY` + `EMAIL_FROM` are set. Without them the verification email never sends and users can't complete signup.

## Test plan

- [ ] CI green (tests + typecheck + lint)
- [ ] Local: hit `/provision` with a new email → 200 accepted, email arrives
- [ ] Local: hit `/provision` with an existing email → 200 accepted (identical body — verify with `diff`)
- [ ] Click verify URL → hit `/redeem` → new email returns token + workspace
- [ ] Click same verify URL again → 410 challenge_invalid_or_expired
- [ ] Hit `/redeem` with existing-user challenge → recovery URL, no token
- [ ] Trip email rate limit (4 provisions in an hour, same address) → 429
- [ ] `install.sh` end-to-end on a fresh macOS shell

## Known follow-ups (out of scope for this PR)

- `apps/api/app/modules/guard/routers/events.py:710` uses the same first-XFF trust anti-pattern. Impact is bounded (authenticated endpoint, IP is display-only), but worth fixing in a follow-up.
- Audit S02 (`ENVIRONMENT=production` in Render) is still pending — conduct-internal #46. That's a dashboard step, not a code change.
- Audit S03 (sandbox LocalSession fallback) is mitigated by S02 landing, but a defense-in-depth fix in `runtime/sandbox_session.py` is still worth doing — conduct-internal #47.

Refs: conduct-internal #45 (closes S01), #46 (S02 unaffected), #47 (S03 mitigated).

### DIFF
--- a/apps/api/app/core/auth.py
+++ b/apps/api/app/core/auth.py
@@ -941,6 +941,26 @@ def resolve_agent_token(token: str, db: Session) -> tuple[str, str] | None:
     return (row[0], row[1]) if row else None
 
 
+def require_platform_operator():
+    """FastAPI dep — 403s unless the caller's Clerk user_id is in
+    settings.platform_operator_clerk_ids. Used to gate cross-tenant ops
+    endpoints (e.g. /guard/trial/ops). A tenant admin role must NEVER
+    be enough on its own — see audit S05.
+    """
+    from app.core.config import settings as _s
+
+    def _dep(user_id: str = Depends(get_user_id)) -> str:
+        # In local/dev mode (no Clerk) user_id == DEV_USER_ID; explicit
+        # opt-in required: set PLATFORM_OPERATOR_CLERK_IDS=dev to enable
+        # ops endpoints on a dev box, otherwise 403 even there.
+        allowed = {s.strip() for s in _s.platform_operator_clerk_ids.split(",") if s.strip()}
+        if user_id and user_id in allowed:
+            return user_id
+        raise HTTPException(status_code=403, detail="platform_operator_only")
+
+    return _dep
+
+
 def token_is_expired(token: str, db: Session) -> bool:
     """True iff token matches a real AgentIdentity row whose expires_at has passed.
 
--- a/apps/api/app/core/config.py
+++ b/apps/api/app/core/config.py
@@ -88,6 +88,18 @@ class Settings(BaseSettings):
     sentry_dsn: str = ""
     app_version: str = "1.0.0"
 
+    # Platform operators — Clerk user IDs that can access cross-tenant ops
+    # endpoints (e.g. /guard/trial/ops). Comma-separated. Empty = nobody,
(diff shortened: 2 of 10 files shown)


sseshachala:
Superseded by #1794 after rebase onto post-#1793 main. Same commits, just replay clean.

Issue #2057, opened 2026-09-17
Governance under load: shared contract for gateway, MCP, and workflows


sseshachala:
## Why this epic exists

Gateway (#2056), MCP tool execution, and workflow execution all do the same governance job: check identity + policy + budget, execute, record. Splitting them into three parallel scaling epics duplicates the contract and lets it drift under pressure.

This epic locks the **shared contract**. Each surface then ships its own scaling work against it. No child stage ships if it violates any invariant below.

## Five invariants — non-negotiable at every stage of every surface

1. **Authorization is fresh enough.** Cached auth may not exceed a bounded staleness window. Revocation is honored within that window even if invalidation messages are lost (Redis Pub/Sub does not guarantee delivery — plan for it).
2. **Budget reservation happens before dispatch.** No paid upstream call fires without a reserved slot in the ledger. Reservation is releasable.
3. **Durable acceptance before upstream dispatch.** Flight Recorder records \"accepted\" durably *before* the outbound call; outcome recorded on completion. Fire-and-forget after dispatch is out of scope.
4. **Retries are idempotent or refused.** An unknown outcome is not permission to retry. Idempotency keys or explicit \"outcome unknown\" states — not blind resend.
5. **Crash-resume is deterministic.** After any worker or instance crash, recorded state is enough to resume correctly, without duplicate charges or duplicate side effects.

## Shared Flight Recorder schema

Every surface writes to the same lineage:

\`\`\`
workspace → agent → (workflow_run) → step → request → attempt
\`\`\`

\`workflow_run\` and \`step\` are nullable for direct gateway/MCP calls. \`attempt\` distinguishes original from retry. This is the join key across surfaces — a workflow that calls Claude, falls back to another model, invokes a tool, and finishes reads as one connected trace.

## Shared enforcement primitives

| Primitive | Consumers |
|---|---|
| Admission control (bounded concurrency + per-workspace caps + workload isolation) | gateway, MCP, workflow workers |
| Auth cache with bounded staleness + version fencing | gateway, MCP |
| Budget ledger (reserve / commit / release) | gateway, MCP |
| Durable audit lifecycle (accepted → finalized) — extends #1959 | gateway, MCP, workflow steps |
| Idempotency keys + \"outcome unknown\" semantics | gateway transport, MCP tools, workflow steps |
| Separate worker pools per surface | ops / deploy |

## Deliverables (this epic ships the contract, not the surfaces)

[...]

Issue #2056, opened 2026-09-17
Gateway hot-path scaling — 4 stages under the governance contract


sseshachala:
Child of #2057 (governance under load — shared contract). Delivers gateway-specific scaling against the five invariants locked in the parent. Any stage that would violate durability, admission, staleness, idempotency, or crash-resume is a blocker — not a scope reduction.

## Discovery — stress test 2026-09-17

100 requests at concurrency 25 against \`cond-6zq8mzpc-claude-sonnet\` via \`scripts/stress_gateway.py\`:

- p50 latency **5.6s → 23.7s** (4× degradation vs baseline).
- **Non-2xx rate: 87%** (52 × 502, 35 × client-side timeout).
- Kill switch fired at request 20.
- Simple UI GETs (\`/organizations\`, \`/preferences\`, \`/projects\`) took **11–40 s** during the incident.
- **Render restarted the API instance.**
- Full recovery took ~60s.

Blast radius: **entire API instance**, not just gateway. Every workspace, every endpoint stalled.

### 1. Sync SQLAlchemy in async handler

\`apps/api/app/modules/guard/gateway_handler.py::handle_gateway_request\` is \`async def\` but calls sync SQLAlchemy:

- \`gateway_handler.py:109\` — \`db.query(_AgentRunToken)\`
- \`gateway_handler.py:120\` — \`db.commit()\`
- \`gateway_handler.py:153\` — \`db.query(_AgentIdentity)\`
- \`gateway_handler.py:156\` — \`db.commit()\`
- \`gateway_handler.py:276\` — \`db.query(_User)\`

Every sync DB call blocks the worker's event loop. Under concurrency 25 across 2 workers, outbound HTTPS handshakes to Anthropic time out → cascading 502s.

### 2. Every request hits the DB on the hot path

Auth check → agent identity → user → policy → vault → durable audit = **~6 DB roundtrips per request**. Serial DB work is a hard throughput ceiling regardless of async.

## 4 stages (refined against the parent contract)

Each stage is independently valuable and shippable. Do NOT bundle. Each has clear stress-test acceptance against a **mock upstream first**, paid canary second.

### Stage 1 — Instrumentation, admission control, blast-radius containment

**Effort: 1–2 days.**

Not just a worker bump. \"More workers alone\" increases DB pressure and does not contain blast radius.

- [ ] Bounded in-flight semaphore on the gateway path.
- [ ] Per-workspace token bucket — one workspace cannot starve others.
[...]

PR #2076, opened 2026-09-17
feat(scaling): PR 6d — atomic budget-reservation ledger (primitive)

sseshachala:
## Summary

Closes the race window in the existing spend-cap path. Today two concurrent requests each read `monthly_cost < cap`, both pass `budget_check`, both proceed, and together they can spend past the cap — because the check-then-write cycle has no reservation between them.

This PR ships the **atomic reservation primitive** — Redis + Lua `check-and-INCRBY` — with kill switch off. No integration into `SpendCapPolicySource` in this PR (deferred to a follow-up, mirroring the 6e→6b split).

## Model

- **committed** — spend already in `guard_audit_events`. Postgres remains source of truth; this PR does not touch the write path.
- **reserved** — in-flight spend estimates in Redis. Counter per `(workspace_id, ai_tool_or_all, YYYY-MM)`. Short-lived — INCRBY at `reserve()`, DECRBY at `release()`.

**Cap enforcement:** `committed + reserved + estimated <= cap`. The check and INCRBY run inside one Lua script so concurrent reserves are serialised by Redis's single-threaded scripting model.

## API

```python
from app.core.budget_ledger import get_budget_ledger, BudgetDecision

ledger = get_budget_ledger()
decision, res = ledger.reserve(
    workspace_id="ws-1", ai_tool="claude",
    estimated_cents=100, cap_cents=10_000, committed_cents=800,
)
if decision == BudgetDecision.ACCEPTED:
    try:
        # ... run the request ...
    finally:
        ledger.release(res)
```

## Kill switch + failure modes

- `BUDGET_LEDGER_ENABLED=false` (default) — nothing calls the primitive; zero prod effect.
- Redis unreachable → `BudgetDecision.REDIS_DOWN`. Callers should fall through to existing `budget_check` (current behaviour).
- `BUDGET_LEDGER_FAIL_CLOSED=true` — override to strict refuse instead of degrade.
- Double-release is safe (Lua clamps counter at 0).
- Zero-cost reservation short-circuits without touching Redis.

## Reconciler

Not required. Redis flush loses in-flight reservations only — committed spend stays in Postgres. In-flight reservations resolve within request lifetime anyway.

## Tests

15 proofs at `tests/core/test_budget_ledger.py`:

- Kill switch off by default; on when flipped
- Reserve within cap → ACCEPTED
- Reserve over cap → EXCEEDED, no side effect
- Reserve respects `committed_cents` (not just raw cap)
- Release refunds reserved
- Double-release clamps at 0
- Zero-cost reservation skips Redis
- Per-tool reservations isolated (F3 cross-tool-bleed invariant)
- Per-workspace reservations isolated
- **50-thread concurrent reserve — exactly N succeed, never more** (the atomicity proof)
- Release during contention frees capacity
- Redis-down → `REDIS_DOWN`, does not raise, does not overshoot
- Stats track outcomes
- Singleton lifecycle

## Requirements

- `fakeredis[lua]==2.38.0` — test-only dep for genuine Lua-script emulation. Real Redis semantics apply: Lua runs single-threaded on the server side, and fakeredis's `lupa`-backed emulator serialises the script body the same way.

## Regression

1921 tests across `tests/core/`, `tests/guard/`, `tests/runtime/` still green.

## Follow-up (separate PR)

- Add `reservation_id` field to `PolicyDecision`
- Wire `SpendCapPolicySource.evaluate()` to call `reserve()` when the flag is on
- Wire `release()` at request end (audit event write is the natural hook)
- Canary validation: run one workspace through the ledger, watch the `reservations_accepted / exceeded / redis_down` counters before global flip

Refs: #2056 (gateway scaling parent), #2057 (governance-under-load).


sseshachala:
Addressed all four findings in 52a4f4e6.

**P1 #1 — Reservation identity via HSET.**
`release()` now looks up the amount from a Redis hash keyed by `reservation_id`, `HDEL`s it, then `DECRBY`s the aggregate counter. Second release for the same id finds the hash empty → no-op. Cannot refund another reservation's slot.
- Test: `test_double_release_does_not_refund_other_reservation` (reviewer's exact 40/60 scenario).

**P1 #2 — Settlement without caller-supplied committed snapshot.**
`reserve()` signature no longer takes `committed_cents`. All three terms (reserved, committed, estimated) are read inside the Lua script — no concurrent commit/release can race between check and increment.
Added `commit(reservation, actual_cents)`: atomically `DECR reserved` / `INCR committed`.
- Test: `test_stale_committed_snapshot_cannot_leak_capacity`, `test_commit_moves_reserved_to_committed`, `test_commit_is_idempotent_by_reservation_id`.

**P1 #3 — Durable reservation log + cold-start reconciler.**
Every `reserve()` writes a `budget_reservations` row *before* touching Redis; `release`/`commit` mark it resolved. On Redis flush or worker cold start, `reconcile()` replays open rows and rebuilds the reserved counter + hash from Postgres, then sets a per-key `ready` flag. `reserve()` returns `NOT_READY` until the flag is present. A flushed Redis cannot silently restore capacity.
New table `budget_reservations` + Alembic migration `0138_budget_reservations.py` (paired per CLAUDE.md).
- Test: `test_reserve_before_reconcile_is_not_ready`, `test_reconciler_restores_reserved_from_durable_log`.

**P2 #4 — TTL preservation on partial release.**
Release Lua now uses `DECRBY` (preserves TTL). When counter reaches zero → `DEL`; when it stays positive → explicit `EXPIRE` to keep the period-scoped TTL alive.
- Test: `test_partial_release_preserves_ttl`.

Also carried over: 50-thread atomicity, Redis-down returns `REDIS_DOWN` + durable row rolled back so no phantom log entries, zero-cost short-circuit.

Kill switch still OFF; wiring into `SpendCapPolicySource` still deferred to a follow-up canary PR.

Regression: 1919 tests green.

PR #2099, opened 2026-09-18
feat(gateway): PR-A2a reserve + settle helpers in gateway_lifecycle

sseshachala:
## Summary
Adds the two pure helpers that own the reserve/settle contract from the corrected design review:

    choose -> policy -> RESERVE -> execute one attempt -> outcome -> SETTLE

Both helpers are **semantics-only** — no upstream IO, no policy eval, no audit writes. **PR-A2b** will call them from ``gateway_handler.py`` with the feature-flag guard. Shipping them dark first so the semantics can be reviewed in isolation before any behavior change lands in production.

## ``reserve_budgets_for_request(...)``

- Looks up applicable budgets via ``lookup_applicable_budgets()`` (#2096).
- Calls ``ledger.reserve_all()`` (#2097) with the full request scope.
- Returns a structured ``ReserveBudgetsResult`` so the gateway handler pattern-matches on outcome without inspecting ``BudgetDecision`` internals.

**Fail-CLOSED baked in — every non-ACCEPTED path must reject:**

| Outcome | Trigger | Caller action |
|---------|---------|---------------|
| ``DISABLED`` | ``BUDGET_LEDGER_ENABLED=false`` | Pre-PR-A behavior; allow dispatch |
| ``ACCEPTED_NO_HARD_CAP`` | No hard-capped budgets apply | Allow dispatch |
| ``ACCEPTED_WITH_RESERVATIONS`` | All applicable budgets accepted | Allow dispatch, hold reservations |
| ``EXCEEDED`` | A budget refused | **Reject** — 402 with refusing budget in reason |
| ``NOT_READY`` | Cold-worker reconciler | **Reject** — 503 |
| ``REDIS_DOWN`` | Ledger unavailable | **Reject** — 503 |
| ``DB_ERROR`` | Invalid input or exception | **Reject** — 503 |

## ``settle_reservations(...)`` truth table

|  ``dispatched``  |  ``actual_cents``  |  Action                    |  Why                                  |
|:-:|:-:|:-|:-|
| ``False`` | any | ``release_all`` | No wire bytes, safe refund |
| ``True``  | int | ``commit_all(N)`` | Success — charge every budget |
| ``True``  | ``None`` | ``PENDING_RECONCILER`` | **Never release after dispatch** — bytes flew, provider may bill; reconciler catches via lease expiry |

The ``dispatched=True + actual_cents=None`` case is the **load-bearing correctness rule**. Releasing then would silently drop real spend.

Empty reservations list -> ``NOOP`` (matches ``ACCEPTED_NO_HARD_CAP``).

## Tests

``tests/guard/test_gateway_reserve_settle.py`` — 14 cases covering every branch of both helpers:

- ``reserve``: DISABLED, invalid ws, no hard caps, accepted, EXCEEDED, NOT_READY, REDIS_DOWN, lookup raise, reserve_all raise.
- ``settle``: empty list, not dispatched, not dispatched with actual_cents (ignored), dispatched+known, **dispatched+unknown** (the load-bearing case).

**14/14 pass locally.** A reviewer can grep the file to prove there is no fail-open branch for hard caps.

Depends on: #2096 (lookup) + #2097 (multi-scope primitives) — both merged.
Feeds into: **PR-A2b** — wire these into ``gateway_handler.py`` at the dispatch site with the feature-flag guard. That is where behavior changes.

## Test plan
- [x] ``pytest apps/api/tests/guard/test_gateway_reserve_settle.py`` — **14/14 pass**.
- [ ] CI: existing gateway + ledger + spend suites remain green (nothing wired yet — dark).


sseshachala:
**HOLD — do not merge.** Reviewer surfaced 9 real issues in the multi-scope arc; #2097 has a load-bearing bug: Redis keys don't include budget/agent/user scope, so a workspace-default and an agent budget both with `ai_tool=NULL` collide on the same counter. Repro: 100¢ request produced 200¢ committed.

Fix 1-4 land as separate PRs against main. This PR will rebase on top once the base is corrected.

PR #2100, opened 2026-09-18
feat(gateway): PR-A2b wire reserve/dispatch/settle into gateway_handler

sseshachala:
## Summary
Wires the multi-scope budget reservation flow into the live gateway request path. **Fail-closed** on any non-ACCEPTED reserve outcome, **dispatch-boundary aware settlement**, and gated behind ``BUDGET_LEDGER_ENABLED`` so **zero behavior change** until ops flips the flag on.

## Four anchor points in ``gateway_handler.py``

**1. Right after ``_open_durable`` succeeds**, before the exception-safe try/finally:

- ``reserve_budgets_for_request()`` with the full request scope (workspace, agent_identity_id, transport, client_tool, clerk_user_id, estimated_cents, request_id).
- On ``EXCEEDED / NOT_READY / REDIS_DOWN / DB_ERROR``: close audit row, return ``budget_block_response()`` (402 / 503).

**2. Right before ``_execute_v2`` fires** (v2 path): set ``_dispatched = True``.
- Any exception past this point is treated as "may have sent bytes" -> settle marks **PENDING_RECONCILER** (never releases; reconciler catches via lease expiry).

**3. Right before ``transport.forward`` fires** (legacy path): same ``_dispatched = True`` flip.

**4. In the exception-safe inner finally**, after ``_close_durable``:
- Derive ``_actual_cents`` from response via ``_extract_token_counts`` + ``_compute_audit_cost`` (same helpers the audit path uses).
- Call ``settle_reservations()`` with reservations list, ``_dispatched``, ``_actual_cents``.
- Truth table:

|  ``_dispatched``  |  ``_actual_cents``  |  Action                     |
|:-:|:-:|:-|
| ``False`` | any | ``release_all`` |
| ``True``  | int | ``commit_all(N)`` |
| ``True``  | ``None`` | **PENDING_RECONCILER** |

Empty list -> NOOP (matches ACCEPTED_NO_HARD_CAP / DISABLED).

## New helpers in ``gateway_lifecycle.py``

- **``estimate_budget_cents(body, provider, model, ai_tool)``** — bounded pre-flight cost estimate. ``input_tokens_approx + output_allowance`` (``max_tokens`` if set, else 4096 default) times per-tool pricing from ``TOOL_PRICING``. Integer cents, **rounded UP** so under-reservation cannot let a request slip past a cap.
- **``budget_block_response(result)``** — 402 for EXCEEDED, 503 for NOT_READY / REDIS_DOWN / DB_ERROR. Response body carries ``reason`` and ``refusing_budget`` for the drawer UI in the follow-up. **Never** includes workspace_id / agent_identity_id (cardinality-safe).

## Tests

``tests/guard/test_gateway_budget_estimate_and_block_response.py`` — **12 cases:**

- ``estimate_budget_cents``: typical request, scales with input length, respects ``max_tokens`` output bound, default allowance, list-of-parts content, defensive on malformed bodies, uses tool pricing (Sonnet-class > Gemini).
- ``budget_block_response``: 402 on EXCEEDED, 503 on NOT_READY / REDIS_DOWN / DB_ERROR, no workspace_id / agent leaks.

**12/12 pass locally. Existing 63 tests across gateway lifecycle, ledger, and audit-durable suites still green — no regressions.**

Depends on: **#2099** (PR-A2a helpers).
Feeds into: PR-B (drawer UI), PR-C (canary flip).

## Explicitly out of scope

- **Streaming path settle timing.** The v2 streaming wrapper handles finalize in its own generator; settle needs to move into that wrapper so reservations clear when the stream drains, not when the handler returns early. Follow-up.
- **Reconciler awareness of new scope columns** (``agent_identity_id``, ``source``, ``client_tool``, ``request_id``). The reconciler already scans open rows; it just needs to know these columns exist for the drawer join. Follow-up.

## Test plan
- [x] ``pytest apps/api/tests/guard/test_gateway_budget_estimate_and_block_response.py`` — **12/12 pass**.
- [x] ``pytest apps/api/tests/guard/test_gateway_reserve_settle.py apps/api/tests/core/test_budget_ledger.py apps/api/tests/guard/test_audit_durable_lifecycle.py apps/api/tests/guard/test_guard_check_prompt_verb.py`` — **63/63 pass** (no regressions in adjacent suites).
- [ ] CI: full test suite green.
- [ ] Post-merge on preview with ``BUDGET_LEDGER_ENABLED=false`` (default): confirm zero behavior change by hitting the gateway and observing no admission changes.
- [ ] PR-C (later): flip ``BUDGET_LEDGER_ENABLED=true`` on one workspace canary, monitor ``/admin/cache-stats`` for ``reservations_accepted`` climbing.


sseshachala:
**HOLD — do not merge.** Depends on #2099 which has correctness issues surfaced in review. Fix PRs coming; this branch rebases after.

Issue #2146, opened 2026-09-19
Contract — all workflow LLM calls MUST go through a gateway profile (CI-enforced)


sseshachala:
## Purpose

Lock in the architecture shipped by #2144 + #2145 so it does not decay:

- No playbook YAML declares a raw model — all use `profile:`
- No code under `apps/api/app/**` instantiates a raw provider SDK client — all LLM traffic flows through the gateway (`/gateway/v1/completions` from #2144)

Without CI enforcement, `model:` and `AnthropicClient(api_key=…)` will creep back in through new playbooks, PRs, and contributor code. This issue is the durable guardrail.

### 1. YAML lint — playbooks + templates + registry

CI check that fails the build when any of these files reference `model:` for LLM configuration:

- `apps/api/playbooks/*.yaml`
- `apps/api/playbook_templates/*.yaml` (or wherever templates live)
- `apps/api/playbooks/registry.yaml`

Allowed key: `profile:` (points at a gateway profile identifier like `cond-e785vpmb-gpt-4o`).

Implementation: `tests/test_contract_playbook_profile_only.py` — walks the YAML tree, asserts no LLM block uses `model:`.

### 2. Python lint — no raw provider clients

CI check that fails when any file under `apps/api/app/**` instantiates a raw provider SDK client.

Blocked constructors / imports:
- `AnthropicClient`, `Anthropic`, `anthropic.Client`, `AsyncAnthropic`
- `OpenAI`, `AsyncOpenAI`, `openai.OpenAI`
- Any equivalent for Perplexity / Google / Groq / OpenRouter / Together

Allowlist (files where the gateway itself needs raw provider clients):
- `apps/api/app/modules/guard/routers/gateway_proxy.py`
- `apps/api/app/modules/guard/gateway_handler.py`
- `apps/api/app/modules/guard/gateway_runtime.py`
- Any file added by #2144 that hosts the `/gateway/v1/completions` implementation

Everything else — playbooks, canvas tools, Lens actors, glens registrations, MCP tools, background jobs, cron jobs, marketplace helpers — MUST call the gateway.

Implementation: `tests/test_contract_no_raw_llm_clients.py` — grep-based check with explicit allowlist. Fails with a clear error message pointing at #2145 for the migration path.

## Scope of the "all workflows" claim

The rule applies to every LLM caller under `apps/api/app/**`:

- Playbook brain / LLM blocks
- Canvas workflow blocks
- Lens tool implementations (chat, mutators, actors)
[...]

Issue #2145, opened 2026-09-19
Epic — Route playbook LLM calls through the gateway (profile-based, zero-secret playbooks)


sseshachala:
## Problem

Playbooks currently call Anthropic (and other providers) **directly** from `apps/api/app/routers/workflows.py`:

- `_resolve_anthropic_key()` reads `ANTHROPIC_API_KEY` from the workspace vault
- A raw `AnthropicClient(api_key=…)` is instantiated per brain-block invocation
- All 39 playbooks in `apps/api/playbooks/registry.yaml` declare `model: claude-*` — a bare model name, not a profile

Consequences:
- Every playbook run needs decrypted vault access
- Vault regressions surface as playbook-run 401s (motivating incident: `run self_driving_network_approval_demo` → `AuthenticationError 401 invalid x-api-key` after vault refactor)
- Playbook LLM calls **bypass** the gateway's guard / policy / reservation / audit flow — different code path from direct-SDK callers
- Model choice is hardcoded at the playbook level instead of centralized in a gateway profile that ops can rotate/change

## Solution

Move playbook LLM calls to a **gateway profile** contract. Playbooks hold zero secrets; the gateway owns credentials, wire format, upstream selection, guard, and audit.

**Before:**
```yaml
- run brain:
    model: claude-sonnet-4-6
```

**After:**
```yaml
- run brain:
    profile: cond-e785vpmb-gpt-4o
```

**Runtime path change** in `workflows.py`:
- Delete `_resolve_anthropic_key()` and the direct `AnthropicClient` construction
- Call `POST /gateway/v1/completions` (from #2144) with `{profile, messages, max_tokens, stream}`
- Gateway resolves profile → creds + wire format + upstream + guard + audit

## Non-goals

- Not deleting `/gateway/v1/{anthropic,openai}/*` — those stay as SDK drop-ins for external users
- Not removing the vault direct-access API — other paths still need it (env vars, integrations, MCP creds)
- No streaming for playbook brain blocks in v1 (matches #2144 v1 scope)

## Scope (in this epic)

1. **Depends on #2144** — profile-native `POST /gateway/v1/completions` shipped
2. Playbook YAML schema: introduce `profile:` on brain / LLM blocks; keep `model:` as deprecated alias for one release with a startup warning
3. `workflows.py`: replace direct Anthropic path with a gateway-completions call; remove `_resolve_anthropic_key()`
[...]

Issue #2144, opened 2026-09-19
Epic — Profile-native gateway endpoint: POST /gateway/v1/completions


sseshachala:
## Problem

Today's `/gateway/v1/{anthropic,openai,perplexity}/...` routes require the client to speak provider-specific wire formats. That's the right shape for SDK drop-ins (`ANTHROPIC_BASE_URL=...`, `OPENAI_BASE_URL=...`).

It is the **wrong** shape for callers written directly against Conduct:
- `scripts/stress_gateway.py`
- `conduct` CLI programmatic paths
- MCP tools
- future Conduct SDK
- curl / marketplace demos
- playbook runtime (see #2145)
- any caller that just wants "post prompt via gateway using this profile"

These callers should NOT need to know whether the upstream is Anthropic or OpenAI. The **profile** is the contract.

## Solution — canonical, profile-native endpoint on the v2 executor

```
POST /gateway/v1/completions
Authorization: Bearer <agent-token>
Content-Type: application/json

{
  "profile": "cond-e785vpmb-gpt-4o",
  "messages": [{"role":"user","content":"..."}],
  "max_tokens": 5,
  "stream": false
}
```

`/completions` is a **new entry point on the existing config-driven Gateway v2 executor**. It does NOT introduce a new routing implementation and it does NOT reuse the legacy `_proxy` from `gateway_proxy.py`.

### Intended flow (identical to what SDK-shaped routes will run once fully migrated to v2)

1. **Authenticate agent** (existing agent-token dependency)
2. **Resolve profile and pin its published config** (`resolve_profile_runtime` in `apps/api/app/modules/guard/gateway_runtime.py`, revision-cached per #2056)
3. **Admission + policy + budget + durable audit** (existing gateway v2 machinery — governance-under-load contract #2057)
4. **Select configured target** from the pinned profile config
5. **Evaluate target-specific policy** (per-target guard rules)
6. **Execute through the target's configured transport** (native SDK / HTTP passthrough / etc — whatever the profile config declares)
7. **Record usage, settle budget, finalize audit**
8. **Return a normalized response** in the canonical envelope

### The only new responsibilities added by `/completions`

- Accept a provider-independent request shape (canonical envelope)
- Convert that request appropriately for each selected target's transport
- Return a consistent normalized response

Everything else — auth, resolution, admission, policy, budget, target selection, target-specific policy, transport execution, usage recording, budget settlement, audit — is the existing Gateway v2 executor, unchanged.

[...]

Issue #1959, opened 2026-09-14
Investigation: Flight Recorder hook vs Gateway gaps, stale client credentials, and durable inference audit design


sseshachala:
## Purpose
Capture the September 14, 2026 investigation and handoff for deeper analysis. This is an evidence-based session summary, not a raw transcript. No credentials, email addresses, prompts, or authentication payloads are included.

## User expectation
Flight Recorder is a core product guarantee. Real Claude/Codex tool calls and Gateway inference must appear in the same activity log, correctly attributed to workspace, agent, and session when supported. Configuration, queued background tasks, and synthetic events must not be presented as proof that real traffic was durably recorded.

## Scope and state
- Local inspection baseline: CLI release branch at 3a555381; installed CLI 0.14.7.
- Fetched origin/main during the latest investigation. No diff in the inspected guard/Gateway modules and CLI guard.py against that baseline at the time of inspection.
- No production inference calls or fabricated audit events were sent during this investigation. Read-only audit/auth requests and Gateway catalog/handshake probes were used.
- No durable Gateway implementation fix has been made yet.
- Local unrelated spend changes and prior UI follow-up changes exist; do not overwrite or mix them into a new fix.

## Timeline and prior work
- PR #1955 fixed repeated-warning suppression incorrectly preventing later tool event submission. Policy warnings may be suppressed in chat without suppressing event capture.
- PR #1956 added authenticated hook-event agent attribution, recorded activity sessions, and session/agent filters, plus responsive UI work.
- CLI 0.14.7 released; version alignment PR #1957. The user installed the release.
- Initially confirmed current-session tool events in the production API, then user screenshots confirmed those rows in Flight Recorder.
- A local UI follow-up refreshes identity/session data on focus, every 15 seconds while relevant tabs are active, and manually. It remains separate uncommitted work: agent-identity/page.tsx, AgentActivitySessions.tsx and associated tests.
- That UI follow-up passed 11 component tests and a browser regression for the first attributed session appearing after focus. These tests do not establish Gateway recording correctness.
- User clarified that model inference and tool calls should share one log. Investigation expanded from hooks to Gateway.
- Tool inspection was intermittently blocked by monthly spend limits of 10 USD and then 20 USD. Later reads succeeded. Exact budget scope/cache behavior was not established; do not assume a specific policy rule was responsible.

## Verified production observations
[...]

PR #2150, opened 2026-09-20
test(gateway): integration coverage for /gateway/v1/completions (follow-up to #2148)

sseshachala:
## Summary

Follow-up to **#2148** (already merged). Adds the end-to-end integration coverage the reviewer flagged as required before claiming governance parity — these two commits were on the source branch when #2148 was squash-merged with only the shim + unit tests, so they land here as a separate PR.

## What's added

**Commit 1 — integration fixture + smoke** (`3d2def01`)

- Reuses the chaos docker-compose (Postgres on non-standard port + Redis); skips cleanly when the two env vars aren't set (no CI break)
- Session-scoped alembic upgrade, autouse per-test truncate of workspaces / agent_identities / gateway_profiles* / guard_audit_events / audit_log / integrations / budget_reservations; Redis flushall; revision + auth cache clear
- Monkeypatches `SessionLocal` to the integration engine (so `handle_gateway_request`'s direct `SessionLocal()` calls hit the integration DB)
- Monkeypatches `NativeHTTPTransport.execute` so outbound HTTP never leaves the process — records the call, returns a canned OpenAI response with target-model rewrite mirrored
- Monkeypatches `resolve_gateway_key` across three import paths so the vault lookup succeeds
- Forces `guard_gateway_profile_v2` + `rollout_pct=100` so cond-... requests don't land on the 501 rollout gate
- `seeded_profile` factory: one workspace + one `cond_api_` token + one published v2 profile with a native_http target
- Smoke test proves the whole stack: publish → POST → 200 → single dispatch captured

**Commit 2 — four parity assertion tests** (`04e29961`)

| # | Test | Invariant |
|---|------|-----------|
| 1 | `test_completions_profile_resolution` | Outbound body carries `target.model` (`openai/gpt-4o`), not caller's cond-alias. Silent fallthrough to legacy v1 routing would leave the alias intact — a match rules that class of regression out. |
| 2 | `test_completions_policy_block_parity` | Same policy BLOCK returns the SAME status + envelope shape via `/gateway/v1/completions` and via `/gateway/v1/openai/v1/chat/completions`. Transport not called from either. |
| 3 | `test_completions_budget_refusal` | EXCEEDED `ReserveBudgetsResult` returns canonical `budget_reservation_refused` envelope with 402/503 AND transport not called. Paid traffic leaking past enforcement is the exact regression this guards. |
| 4 | `test_completions_audit_finalization` | Resulting `guard_audit_events` row carries `route=/gateway/v1/completions`, `routing_meta.v2_operation=openai_chat_completions`, `routing_meta.revision_id=<pinned>`, `agent_identity_id=<seeded token's identity>`, `decision=allowed`. #1959 Phase 0 attribution invariant. |

## Design notes

- Policy + budget seeding is monkeypatched at the evaluator / reservation seams rather than seeded as real rows. Those layers already have their own end-to-end coverage under `tests/chaos` and `tests/guard/`. What this suite locks is that `/completions` produces the **same downstream behavior** as the SDK-shaped route — the parity contract, not the source-of-truth logic.
- Stub transport mirrors the real `NativeHTTPTransport` model rewrite (`payload["model"] = target.model`) so tests assert against the wire-level body, not the pre-rewrite payload.
- `guard_audit_events` (Guard's own table) is on the truncate list. The `audit_log` model in `app/models/` is platform-level and is not written by the gateway path.

## Test plan

- [x] 5 integration tests pass locally against the chaos containers in ~2.3s
- [x] 70 adjacent unit tests still pass (`tests/guard/test_gateway_completions_endpoint.py` + `test_gateway_proxy_routes.py` + `test_gateway_handler_v2_wiring.py`)
- [ ] Verify chaos docker-compose stays available in CI env (or document that these skip in CI)

## Iteration landmines documented in commit 1

For anyone extending this suite later — six failure cycles to reach green, notable:

1. `cond_agt_` needs a `guard_member_config` link; `cond_api_` doesn't → use `cond_api_` for machine tokens
2. `profiles.active_revision_id` ↔ `revisions.profile_id` FK is circular → 2-step insert (profile with NULL, revision, then UPDATE profile)
3. v2 uses `NativeHTTPTransport.execute`, NOT the v1 `ProviderTransportRegistry` → stub the right one
4. `guard_gateway_profile_v2` rollout gate returns 501 unless force-enabled in the fixture
5. v2 finalize writes `execution_status='ok'` (v1 uses `'success'`) — parity tests accept both
6. Policy blocks land as HTTP 403 with the fail-closed envelope shape, not 451

## Related

- Follow-up to #2148 (merged)
- Epic: #2144
- Blocks: #2145 (playbook migration), #2146 (CI contract)


sseshachala:
Superseded by #2149 (opened by @sseshachala at the same time and containing the same three commits). Closing to avoid duplication.

PR #2179, opened 2026-09-21
fix(gateway): revert anthropic-version bump — 2024-10-22 not a valid header

sseshachala:
## Symptom

After #2177 shipped \`anthropic-version: 2024-10-22\`, every Anthropic request 400'd. Broke ALL Claude traffic in prod, not just vision.

## Root cause

I invented \`2024-10-22\` speculatively. Anthropic returns 400 "invalid version" for any date that isn't in their published list — they don't cut new version headers frequently. Stable current is \`2023-06-01\` and they add features backward-compatibly under it (URL-source images, computer-use, prompt caching, extended thinking all shipped WITHOUT a new version header).

## Proof

Local test with a synthetic 1x1 PNG data URL to Claude Sonnet 4.6:
- Server default \`2024-10-22\` → 400
- Override via client header \`anthropic-version: 2023-06-01\` → 200 with real description ("orange color")

## Fix

Revert to \`2023-06-01\`. Added warning comment: DO NOT invent version dates. Prefer \`anthropic-beta\` for feature-flagged behavior.

## URL-source image problem remains

Even at 2023-06-01, \`{type:"image",source:{type:"url",url:"..."}}\` gets 400. That was the ORIGINAL problem. The fetch+base64 fallback in #2177 is the real fix — this revert just stops us breaking every OTHER Anthropic request while the fallback handles vision.

## Test plan
- [x] Local override test proves 2023-06-01 accepts base64 image; 2024-10-22 rejects
- [ ] CI green
- [ ] Live retry once deployed — expect all Anthropic requests healthy again


sseshachala:
Cherry-picked revert commit into #2178 per user request. Closing this PR.

Issue #2170, opened 2026-09-21
story: route brain_block through the Conduct Gateway (currently calls providers directly)


sseshachala:
Discovered during #2158 completion work.

## Today

\`apps/api/app/runtime/blocks/brain_block.py\` calls provider APIs (OpenAI, Anthropic) through \`app.runtime.adapters.*\` — bypassing Conduct's own Gateway (\`/gateway/v1/completions\`).

Consequences:

- Brain-block LLM calls don't inherit Gateway governance: policy checks, redaction, admission, tool_call response gate, tool_call correlation IDs.
- Two code paths to reason about — direct-provider and gateway.
- Any Gateway-only feature (streaming buffered-delta validation, per-profile capability gates, tool_call correlation, vision) is invisible to brain_block until this is routed.

## Proposal

Add a runtime transport option so brain_block can dispatch its completion calls through \`/gateway/v1/completions\` when configured. Existing direct-provider path stays as fallback / for tests.

## Scope

- New adapter path: \`GatewayHTTPTransport\` that hits the local Gateway with a canonical OpenAI-shape request and returns a normalized \`LLMResponse\`.
- Config flag / block param to opt-in per workspace or per block.
- Wire the \`X-Conduct-Tool-Correlation-Ids\` header into the outbound \`run_events\` entry for each tool execution.

## Out of scope

- Retiring the direct-provider adapters (they're used for tests and edge cases).
- Any provider surface not already handled by the Gateway.

PR #2181, opened 2026-09-21
feat(workflows): pin published Gateway profile per workflow (#2170 PR 1)

sseshachala:
## Summary

PR 1 of the #2170 epic. Adds a per-workflow pin to a published Gateway profile so brain_block (PR 2) can route through it at runtime.

- Migration `0146_workflows_gateway_profile_id`: nullable UUID FK on `workflows` → `gateway_profiles(id)`, `ON DELETE SET NULL`, named index.
- Model: `Workflow.gateway_profile_id` with explicit constraint + index names (matches repo convention).
- Schema: `WorkflowUpdate.gateway_profile_id` + `WorkflowOut.gateway_profile_id`.
- Router `PUT /workflows/{id}` validates the pin at the boundary:
  - draft profile (`active_revision_id IS NULL`) → 400
  - profile in another workspace → 404 (no existence disclosure)
  - explicit `null` clears the pin without a lookup
- Frontend: Gateway profile card in `WorkflowSettingsPanel`, populated from `guard.gatewayProfilesV2.list(...)` filtered to `active_revision_id != null`.

Nothing consumes the pinned value yet — this is the picker only. brain_block cut-over lands in PR 2 of the epic.

## Test plan

- [x] `pytest tests/test_workflow_crud.py` — 20/20 pass (4 new tests: published-accepted, draft-400, cross-workspace-404, clear-to-null)
- [x] `tsc --noEmit` clean on `apps/web`
- [ ] CI: `test_alembic_check_no_schema_drift` (runs against real DB)
- [ ] Manual: open a workflow's Settings → dropdown lists only published profiles → pick one → refresh, still selected


sseshachala:
Reviewer P1 addressed (item 5): `body.template` audit-call bug fixed. Endpoint now returns 200 on settings-only saves — the Gateway profile dropdown reports success and retains selections properly. All 4 new tests updated to assert HTTP 200 instead of working around the 500.

Rebased #2182 on top of this so the whole train is clean.

Ready to re-review.

Issue #2155, opened 2026-09-20
Tools follow-up — Streaming + tools with buffered-delta validation


sseshachala:
Deferred from the Tools / function calling epic. Fill in the epic number after it's filed.

## Problem

The initial Tools PR rejects `tools + stream=true` at the shim because OpenAI streams `delta.tool_calls[].function.arguments` as fragments across many SSE chunks — a secret can span deltas. A per-chunk redactor sees neither the start nor the end of a match. A response-gate scan on a completed body cannot recall bytes already delivered to the client.

Result today: streaming tools = 400 at the shim with a message pointing at this issue.

## Scope

- Buffer each tool call's argument fragments in the stream wrapper until `finish_reason: "tool_calls"` (or the tool_call's index closes).
- Validate + redact the assembled arguments (same rules as non-streaming: parse JSON → redact string leaves → serialize).
- On validation failure — refuse to emit that tool_call frame at all. Downstream consumer sees the assistant's text content (if any) and a synthetic error frame naming the reason.
- Non-tool text deltas continue to stream chunk-by-chunk unchanged. Only tool_call arg fragments are held.
- Backpressure implications: admission slot held longer, TTFB unaffected for text, TTFB for tool-heavy prompts effectively equals total generation time. Document explicitly.

## Constraints

- Zero raw unsafe bytes reach the client. Ever.
- 451 (policy) and 502 (`tool_arguments_validation_failed`) semantics from non-streaming apply identically once buffered.
- Coordinator must NOT retry-next-target on either terminal outcome (same as non-streaming).
- Response-gate reason recorded on the audit row.

## Out of scope

- Buffered redaction for non-OpenAI tool shapes (Anthropic `tool_use` etc.) — those wait on their own target-support epic.

## Dependencies

- Non-streaming tools support must ship first (parent epic).

sseshachala:
Parent epic: #2159.

PR #2182, opened 2026-09-21
feat(runtime): brain_block routes via pinned Gateway profile (#2170 PR 2)

sseshachala:
## Summary

PR 2 of the #2170 epic. Depends on #2181 (PR 1). When a workflow has `gateway_profile_id` pinned, `brain_block` dispatches through the new canonical Gateway route instead of the direct-provider adapters.

- New adapter `apps/api/app/runtime/adapters/gateway_profile.py` — `GatewayProfileClient` speaks the canonical envelope (`{profile, messages, tools, max_tokens}`) to `POST {conduct_proxy_url}/completions`. Mirrors OpenAIClient's public API so `brain_block` doesn't branch beyond the constructor.
- `LLMResponse.correlation_ids: dict[str, str]` — populated only by the new client, empty for direct-provider adapters. `brain_block` emits each tool_call's correlation id into its per-tool `run_events` entry.
- `brain_block` — reads `workflow.gateway_profile_id`, looks up the profile, requires a non-null `active_revision_id`, and swaps the client at instantiation time. Direct-provider path unchanged when the workflow has no profile pinned.

Anthropic + OpenAI target profiles both work day one — the profile picks the target, the gateway handles the shape.

**Streaming is deferred to a follow-up.** The SSE reassembly of `delta.tool_calls[].function.arguments` needs the same buffered-delta contract the shim wraps upstream (see #2155), and shipping it here doubles the diff without unlocking a customer feature. `GatewayProfileClient.stream()` raises `NotImplementedError`; brain_block's agentic loop uses `create()` only, so nothing calls it. Direct-provider adapters keep their streaming paths in the meantime.

## Depends on

- [ ] #2181 — PR 1 (workflow → profile picker + schema)

Merge #2181 first, then rebase this branch onto main before merging.

## Test plan

- [x] `pytest tests/runtime/test_gateway_profile_client.py` — 8/8 pass
  - canonical envelope (profile vs model, URL, auth passthrough)
  - tools schema conversion (`input_schema` → `function.parameters`)
  - tool_calls → LLMToolUseBlock
  - correlation header parses onto `correlation_ids`
  - missing header → empty dict
  - 4xx surfaces
  - `make_assistant_turn` / `make_tool_results_turn` parity with OpenAIClient
- [x] `pytest tests/test_workflow_crud.py tests/test_brain_block_config_parity.py` — 30/30 pass (no regressions)
- [ ] Manual soak: pick a workflow, set gateway profile in Settings, run it, confirm audit row lands and `run_events` carry `tool_call_correlation_id`


sseshachala:
All four reviewer P1s addressed. Rebased on top of updated #2181.

**1. Fail-closed on assigned-but-unavailable profile** (`brain_block.py`) — profile lookup now RAISES when the row is missing, has no `active_revision_id`, or hits a lookup exception. Only truly-unassigned workflows (NULL) use legacy routing. Test: `test_assigned_unpublished_profile_fails_closed`, `test_assigned_missing_profile_fails_closed`.

**2. Cost accounting from response `model` + `usage`** (`gateway_profile.py`) — adapter takes `pricing_snapshot` and computes cost via the same pricing table the direct-provider adapters use. `_infer_provider_from_model` covers every prefix in the shipped pricing table (Anthropic/OpenAI/Perplexity/Together); unknown prefixes return 0 (safer than misattributing, gateway audit row has the authoritative value). Test: `test_create_computes_cost_for_known_provider_prefix`, `test_create_zero_cost_for_unknown_model_prefix`, `test_infer_provider_from_model_covers_shipped_prefixes`. If a future model prefix escapes the map, follow-up is a gateway-side `X-Conduct-Attempt-Cost-Usd` response header.

**3. Profile resolution runs before `MissingProviderKey`** (`brain_block.py`) — provider-key check is now conditional on `not _profile_assigned`. Workflows authenticating entirely through the Gateway profile no longer need a legacy vendor key on the block. Test: `test_profile_routed_workflow_skips_legacy_provider_key_check`.

**4. Adapter forces `max_attempts=1`** (`gateway_profile.py`) — the gateway owns retry policy on the profile. A terminal `conduct_gateway_tool_arguments_validation_failed` 502 is an already-paid attempt; adapter-level retries would triple the paid inference cost. Test: `test_create_forces_single_attempt`.

**Tests**: 45/45 pass across the touched files (`test_workflow_crud` + `test_brain_block_config_parity` + `test_gateway_profile_client` + `test_brain_block_gateway_profile_routing`). 7 new cases specifically for the four fixes.

Ready to re-review.

PR #2183, opened 2026-09-21
feat(runtime): brain_block streams through canonical /completions (#2170 PR 3)

sseshachala:
## Summary

PR 3 of the #2170 epic. Adds SSE reassembly to `GatewayProfileClient` so `brain_block` can stream through the canonical `/completions` shim without changing its agentic loop.

- Reassembles text deltas + tool-call fragments (`choices[0].delta.tool_calls[].index` merge) + final usage chunk into the same OpenAI chat-shape body the non-streaming branch consumes.
- Correlation header (`X-Conduct-Tool-Correlation-Ids`) read off streaming response headers before body consumption — per-tool_call correlation still lands in `run_events`.
- Opt-in behind two flags: gateway-side `guard_gateway_tools_stream_enabled` (existing, #2155 buffered-delta validation) + new runtime `guard_brain_streaming_enabled`. Both must be on. Default OFF.

Ships alongside PR 4 (retire legacy path).

## Test plan

- [x] `pytest tests/runtime/test_gateway_profile_client.py` — 18/18 pass (6 new streaming cases)
- [x] `pytest tests/runtime/test_brain_block_gateway_profile_routing.py tests/test_workflow_crud.py tests/test_brain_block_config_parity.py` — 33/33 pass (no regressions across touched files)
- [ ] Manual soak once ops flips both flags: run a workflow with tools + profile pinned, confirm audit row + `run_events` correlation still land


sseshachala:
All 5 reviewer findings addressed.

**P1 — client no longer sends `stream_options`** (`gateway_profile.py`). Canonical shim rejects that field (`extra_forbidden`); the shim injects `stream_options.include_usage: true` server-side when `stream=true`. Test: `test_stream_does_not_send_stream_options_client_side` — asserts the outbound payload has no `stream_options` key.

**P1 — SSE `error` events raise `GatewayStreamError`** (`gateway_profile.py`). Any top-level `{"error": {...}}` frame is terminal; the runtime never treats it as a completed turn. Test: `test_stream_error_sse_event_raises_terminal` — reproduces the reviewer's probe (`conduct_gateway_tool_arguments_validation_failed` error frame followed by `[DONE]`).

**P1 — incomplete streams raise `GatewayStreamError`** (`gateway_profile.py`). Neither `[DONE]` nor any `finish_reason` → raise. Also validates each reassembled `tool_call.arguments` string parses as JSON before returning — malformed → raise, never dispatch. Tests: `test_stream_incomplete_no_done_no_finish_raises`, `test_stream_invalid_tool_call_arguments_raises_before_dispatch`.

**P2 — correlation IDs via SSE `conduct` frames** (`gateway_profile.py`). Reassembler parses `{"conduct": {"tool_call_correlation_ids": {...}}}` frames and unions with the header value. Streaming responses can drop custom headers at intermediary hops, so SSE is the primary path there. Tests: `test_stream_correlation_ids_via_sse_event`, `test_stream_correlation_ids_merges_sse_and_header`.

**P2 — per-profile streaming safety** (`brain_block.py`). brain_block now loads the published revision's target list and sets `stream_enabled=True` only when global flag is on AND every target is streaming-safe (no Anthropic). Canonical→Anthropic still 501s at the gateway; global flag on shouldn't force streaming for Anthropic-target profiles. Also introduces `GatewayStreamError` export so dag_runner/audit can distinguish terminal stream failures from ordinary LLM exceptions. Tests: `test_anthropic_target_profile_disables_streaming_even_when_flag_on`, `test_openai_only_profile_streams_when_flag_on`.

**Tests**: 62/62 pass across `test_gateway_profile_client` + `test_brain_block_gateway_profile_routing` + `test_workflow_crud` + `test_brain_block_config_parity`. 8 new cases specifically for these fixes.

Ready to re-review.

